PR #143, opened 2023-03-29
fix: Change ObjectMeta value in Ingress analyser.

arbreezy:
## 📑 Description
A tiny PR to fix the Object meta value we pass in GetParent() for Ingress analyzer

## ✅ Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

PR #144, opened 2023-03-29
fix: ingress object meta value

matthisholleville:
Closes # 

## 📑 Description

A merge request that includes the modifications made in [arbreezy](https://github.com/arbreezy) merge request available https://github.com/k8sgpt-ai/k8sgpt/pull/143 and adds Ingress to the switch case GetParents.

## ✅ Checks

### DIFF
--- a/pkg/analyzer/ingressAnalyzer.go
+++ b/pkg/analyzer/ingressAnalyzer.go
@@ -55,7 +55,7 @@ func AnalyzeIngress(ctx context.Context, config *AnalysisConfiguration, client *
 			Error: value.FailureDetails,
 		}
 
-		parent, _ := util.GetParent(client, value.Endpoint.ObjectMeta)
+		parent, _ := util.GetParent(client, value.Ingress.ObjectMeta)
 		currentAnalysis.ParentObject = parent
 		*analysisResults = append(*analysisResults, currentAnalysis)
 	}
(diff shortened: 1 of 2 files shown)


arbreezy:
@matthisholleville looks good, no worries.
I don't see the output of a missing secret when I analyze the Ingress object while having more errors for the same Ingress object, curious if you can replicate it as well.

matthisholleville:
Can you run : `k8sgpt analyze --namespace test-ci-new-repo -o json | jq .` please ?

This is my output : 

```
...
{
  "kind": "Ingress",
  "name": "my-namespace/my-ingress",
  "error": [
    "Ingress uses the service my-namespace/my-ingress which does not exist.",
    "Ingress uses the secret my-namespace/my-missing-secret as a TLS certificate which does not exist."
  ],
  "details": "",
  "parentObject": "xxxxxx"
}
```

Do you have both errors?

matthisholleville:
@arbreezy You are right that if you don't use the JSON output format, you only get the first error (regardless of the analyzer). I don't think it's related to ingress exclusively. We should open an issue?

arbreezy:
Yes with json output it looks good and code is fine, not sure why plain text isn't showing it.
Is it working for you with both output formats?

arbreezy:
I already touched base on slack, going to have a proper look tomorrow as well, thanks for the prompt response @matthisholleville , most likely we need to open a new issue.

matthisholleville:
@arbreezy Indeed, only the first error from the array is displayed : https://github.com/k8sgpt-ai/k8sgpt/blob/main/cmd/analyze/analyze.go#L144

AlexsJones:
Please check the bug fix I have proposed for the output and if it fixes your issue we can get this in also

matthisholleville:
All errors are displaying correctly with your fix. Great job!

AlexsJones (approved):

PR #184, opened 2023-04-02
refactor: first attempt with more interfaces

thschue:


pkg/analyzer/analyzer.go
```diff
@@ -2,136 +2,157 @@ package analyzer
...
-	// use active_filters
-	for _, filter := range activeFilters {
-		if analyzer, ok := analyzerMap[filter]; ok {
-			if err := analyzer(ctx, config, client, aiClient, analysisResults); err != nil {
-				return err
-			}
-		}
+	analyzerConfig.PreAnalysis = make(map[string]common.PreAnalysis)
+
+	switch analyzer {
```
AlexsJones, line 54: I personally would have added a `GetName()` on the interface to avoid a switch case here

PR #200, opened 2023-04-03
chore: analyzer and ai interfacing

thschue:
## 📑 Description
* Create Interface for AI providers
* Move analyzers to folders
* Restructure analyzers a bit
* Separate analyzers and analysis

## ✅ Checks

- [X] My code requires changes to the documentation
- [X] I have updated the documentation as required


cmd/analyze/analyze.go
```diff
@@ -53,50 +54,44 @@ var AnalyzeCmd = &cobra.Command{
...
-		switch backendType {
-		case "openai":
-			aiClient = &ai.OpenAIClient{}
-			if err := aiClient.Configure(token, language); err != nil {
-				color.Red("Error: %v", err)
-				os.Exit(1)
-			}
-		default:
-			color.Red("Backend not supported")
+		aiClient := ai.NewClient("openai")
```
matthisholleville, line 57: Shouldn't we keep the backend variable here to maintain consistency with the rest of the code? From what I understand, regardless of the value sent in the backend variable, we currently return `openai`. I believe this will make the code easier to modify when we add a new backend.

pkg/ai/iai.go
```diff
@@ -1,8 +1,20 @@
...
+	"context"
+)
 
 type IAI interface {
 	Configure(token string, language string) error
 	GetCompletion(ctx context.Context, prompt string) (string, error)
+	Parse(ctx context.Context, prompt []string, nocache bool) (string, error)
+}
+
+func NewClient(provider string) IAI {
```
matthisholleville, line 13: Should the client's choice be located at the interface level? Shouldn't we move it to the client file instead?

cmd/analyze/analyze.go
thschue: Thank you, yes, changed it :-)

pkg/ai/iai.go
thschue: As it is decided here which client will be created, I think it's good to have it here ... But to be honest, I'm still looking for a way to make this cleaner :-)

thschue:
> I like the signature change of the analyzers but I personally dislike the foldering of each analyzer, it feels unnecessary without evident gain

I moved the analyzers back to a single folder.

pkg/ai/iai.go
matthisholleville: Personally, I prefer having my client declaration at the same level as my methods (similar to how Vault does it with its [client](https://github.com/hashicorp/vault/blob/main/api/client.go#L584)). But it's a subjective choice that doesn't block the validation of this code 😄

AlexsJones (approved):

matthisholleville (approved):

PR #212, opened 2023-04-04
feat!: add support for new configuration format

### LINKED ISSUE #167: Break out the settings from the openAI client into config
We need to have at least the default model as a parameter that is in the config for long term support of models like davinci via openai
e.g.
```
ai:
 providers:
   - name: openai
      model: GPT3-turbo
```

matthisholleville:
Closes #167

## 📑 Description

This commit adds support for a new configuration format that is not backwards compatible with the previous format. This is a breaking change and requires users to update their configuration files to use the new format.

BREAKING CHANGE: The format of the configuration file has changed. Users must update their configuration files to use the new format.

## ✅ Checks

## ℹ Additional Information

Last configuration format : 

```yaml
backend_type: openai
openai_key: xxxxxx
```

New configuration format : 

```yaml
ai:
    providers:
        - name: openai
          model: gpt-3.5-turbo
          password: xxxxxx
```

If the configuration format is incorrect : 

```bash
➜  k8sgpt git:(feature/config-new-format) ✗ ./k8sgpt analyze --namespace k8sgpt --explain --no-cache
Error: AI provider not specified in configuration. Please run k8sgpt auth
```

If the provider name is not configured : 

```bash
➜  k8sgpt git:(feature/config-new-format) ✗ ./k8sgpt analyze --namespace k8sgpt --explain --no-cache --backend fake
Error: AI provider fake not specified in configuration. Please run k8sgpt auth
```

### DIFF
--- a/cmd/analyze/analyze.go
+++ b/cmd/analyze/analyze.go
@@ -35,29 +35,37 @@ var AnalyzeCmd = &cobra.Command{
 	provide you with a list of issues that need to be resolved`,
 	Run: func(cmd *cobra.Command, args []string) {
 
-		// get backend from file
-		backendType := viper.GetString("backend_type")
-		if backendType == "" {
-			color.Red("No backend set. Please run k8sgpt auth")
+		// get ai configuration
+		var configAI ai.AIConfiguration
+		err := viper.UnmarshalKey("ai", &configAI)
+		if err != nil {
+			color.Red("Error: %v", err)
 			os.Exit(1)
 		}
-		// override the default backend if a flag is provided
-		if backend != "" {
-			backendType = backend
+
+		if len(configAI.Providers) == 0 {
+			color.Red("Error: AI provider not specified in configuration. Please run k8sgpt auth")
+			os.Exit(1)
 		}
-		// get the token with viper
-		token := viper.GetString(fmt.Sprintf("%s_key", backendType))
-		// check if nil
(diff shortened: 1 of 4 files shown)


thschue:
Hello @matthisholleville! What should users do to transition from the old to the new format? I think we should add some information and/or automatically transition from the old to the new format, if possible, somehow. WDYT?

matthisholleville:
Users will need to relaunch the command k8sgpt auth as indicated by the following error message.

```bash
➜  k8sgpt git:(feature/config-new-format) ✗ ./k8sgpt analyze --namespace k8sgpt --explain --no-cache
Error: AI provider not specified in configuration. Please run k8sgpt auth
```

I will think about how to make this transition simpler for the user, but I feel that adding code to manage the transition from the old to the new format will add complexity to the code that I do not necessarily find useful. What do you think?

matthisholleville:
Users will need to relaunch the command k8sgpt auth as indicated by the following error message.

```bash
➜  k8sgpt git:(feature/config-new-format) ✗ ./k8sgpt analyze --namespace k8sgpt --explain --no-cache
Error: AI provider not specified in configuration. Please run k8sgpt auth
```

I will consider how to make this transition simpler for the user. However, I am concerned that adding code to manage the transition from the old to the new format may introduce unnecessary complexity to the codebase. What do you think?

Issue #216, opened 2023-04-05
Incorrect findings for HorizontalPodAutoscaler


sybo-matthew:
Checklist:

* [x] I've searched for similar issues and couldn't find anything matching
* [x] I've included steps to reproduce the bug.
* [x] I've included the version of Kubernetes and k8sgpt.

### Subject of the issue

k8sgpt returns incorrect information reporting missing information about HorizontalPodAutoscaler. Specifically: `- Error: HorizontalPodAutoscaler uses Deployment/teams as ScaleTargetRef which does not exist.` 

### Your environment

* Version of Kubernetes
Client Version: v1.24.12-dispatcher
Kustomize Version: v4.5.4
Server Version: v1.24.9-gke.3200
* MacOS 13.2.1
* Version of k8sgpt
k8sgpt version 0.1.8

### Steps to reproduce

* Run `k8sgpt analyze --explain`
* Receive output of: `- Error: HorizontalPodAutoscaler uses Deployment/teams as ScaleTargetRef which does not exist.`
* Confirm deployment does exist in namespace. 
``` kubectl get deployment teams -n other-namespace
NAME    READY   UP-TO-DATE   AVAILABLE   AGE
teams   2/2     2            2           144d
```
* Confirm autoscaling is working
`Name:                                                  teams
Namespace:                                          other-namespace
Annotations:                                           <none>
CreationTimestamp:                              Fri, 11 Nov 2022 11:07:40 +0100
Reference:                                             Deployment/teams
Metrics:                                               ( current / target )
  resource cpu on pods  (as a percentage of request):  1% (2m) / 80%
Min replicas:                                          2
Max replicas:                                          150
Deployment pods:                                       2 current / 2 desired
Conditions:
  Type            Status  Reason            Message
  ----            ------  ------            -------
  AbleToScale     True    ReadyForNewScale  recommended size matches current size
  ScalingActive   True    ValidMetricFound  the HPA was able to successfully calculate a replica count from cpu resource utilization (percentage of request)
  ScalingLimited  True    TooFewReplicas    the desired replica count is less than the minimum replica count
Events:           <none>`

### Expected behaviour
k8sgpt should not be trying to check the HPA at all. 

`❯ k8sgpt filters list
Active:
> Pod
> ReplicaSet
> PersistentVolumeClaim
> Service
> Ingress
Unused:
> HorizontalPodAutoScaler
> PodDisruptionBudget`

### Actual behaviour
[...]

PR #230, opened 2023-04-09
fix: use hpa namespace instead analyzer namespace

### LINKED ISSUE #229: User experience around the optional filter
Checklist:

---

### Subject of the issue

So I ran `k8sgpt analyze` and this is the result:

```text
- Error: HorizontalPodAutoscaler uses StatefulSet/keycloak as ScaleTargetRef which does not exist.
```

So I think to myself, let me narrow the scope and only analyze the namespace wherein this was found to be an issue.

`k8sgpt analyze --filter HorizontalPodAutoscaler --namespace NAME_OF_THE_NAMESPACE`

and I get

```text
No problems detected
```

I'm like what?

Then I find that the HPA filter is optional. I therefore enable it by executing `k8sgpt filters add HotizontalPodAutoScaler`. While doing this I noticed that I had mistyped the HotizontalPodAutoScaler filter ( I had written **HotizontalPodAutoscaler** ).

But, still when executing `k8sgpt analyze --filter HorizontalPodAutoScaler --namespace NAME_OF_THE_NAMESPACE` I get `No problems detected`.

---

### Your environment

* Kubernetes v1.24.9+k3s1
* K8sgpt v0.2.0
* kubectl v1.26.3
* WSL Ubuntu 20.04

---

### Behaviour description

It would be great if the k8sgpt tool:

- Reported `unknown` filter when executing `k8sgpt analyze` commands and the filter is misspelled or otherwise not found, instead of the confusing `No problems detected` message. When one just ran `k8sgpt analyze` for all namespaces and a problem with one or more `HorizontalPodAutoScaler's` where reported.

- That using the `--namespace` option actually worked when using `--filter HorizontalPodAutoScaler`. Where as now I "just" get `No problems detected` when using the `--namespace` parameter. When not using it a problem with one or more `HorizontalPodAutoScaler's` is reported. 

---

### Additional context / screenshots

I also agree with the issue description in [#216 ](https://github.com/k8sgpt-ai/k8sgpt/issues/216) saying

```text
k8sgpt should not be trying to check the HPA at all.
```

As the HPA was also reported on even when not enabled. That's how I first bumped into the scenario I'm seeing.

---

Thank you very much.

matthisholleville:
Closes #229 #216

## 📑 Description

Fixed a bug that did not retrieve the correct namespace when analyzing HPA.

## ✅ Checks

### DIFF
--- a/pkg/analyzer/hpa.go
+++ b/pkg/analyzer/hpa.go
@@ -2,6 +2,7 @@ package analyzer
 
 import (
 	"fmt"
+
 	"github.com/k8sgpt-ai/k8sgpt/pkg/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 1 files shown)


thschue (approved):
lgtm

PR #254, opened 2023-04-12
feat: http server

### LINKED ISSUE #183: feature: provide api for k8sgpt
Checklist:
* [X] I've discussed this feature in the #k8sgpt slack channel

* [ ] Yes
* [X] No

Sometimes, it's easier to integrate in other tools via a webservice than a CLI and therefore, it might make sense to provide a simple way to execute k8sgpt via an HTTP request.

Created a small prototype here:

serve

The code is available in the branch feat/serve.

* There would be an option to run k8sgpt in-cluster and let tools consume it
* As a result, integrations for K8sGPT might get very easy to write

## Potential drawbacks
* Security, we might need some authorization
* The k8sgpt binary gets a bit larger (~52M instead of 39M)

thschue:
Closes #183 

## 📑 Description
* Adds `k8sgpt serve` command
* Introduces NewAnalysis function to unify API and cmd operations
* Fixes caching issues with multiple providers

## ✅ Checks


cmd/serve/serve.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		token := viper.GetString(fmt.Sprintf("%s_key", backendType))
+		// check if nil
+		if token == "" {
+			color.Red("No %s key set. Please run k8sgpt auth", backendType)
+			os.Exit(1)
+		}
+
+		server := server2.K8sGPTServer{
```
AlexsJones, line 41: Could we rename the package import from `server2` to something like `serverProvider` ?

pkg/server/main.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+	"github.com/fatih/color"
+	"github.com/k8sgpt-ai/k8sgpt/pkg/analysis"
+	"net/http"
+	"os"
+)
+
+type K8sGPTServer struct {
```
AlexsJones, line 11: I think renaming this to `Configuration` makes more sense as we don't have a ref to the actual `http.Serve` inside of this object

pkg/server/main.go
```diff
@@ -0,0 +1,74 @@
...
+	color.Green("Starting server on port " + s.Port)
+	err := http.ListenAndServe(":"+s.Port, nil)
+	if err != nil {
+		fmt.Printf("error starting server: %s\n", err)
+		return err
+	}
+	return nil
+}
+
+func getBoolParam(param string) bool {
```
matthisholleville, line 69: Shouldn't we try to convert param to bool and return the result? 

```go
func getBoolParam(param string) bool {
	b, err := strconv.ParseBool(param)
	if err != nil {
		// Handle error if conversion fails
		return false
	}
	return b
}
```

With this modification, even if the user sends "True" the function will interpret the boolean correctly, whereas with the current version, the function will return `false`.

Example:

```go
func getBoolParam(param string) bool {
	b, err := strconv.ParseBool(param)
	if err != nil {
		// Handle error if conversion fails
		return false
	}
	return b
}

func oldGetBoolParam(param string) bool {
	if param == "true" {
		return true
	}
	return false
}

func main() {
	test := "True"
	fmt.Println(getBoolParam(test))
	fmt.Println(oldGetBoolParam(test))
}

---------

go run main.go
true
false
```

cmd/serve/serve.go
thschue: fixed this

pkg/server/main.go
thschue: changed this

pkg/server/main.go
AlexsJones: +1 on this, 

we could toLower + convert bool

pkg/server/main.go
```diff
@@ -0,0 +1,74 @@
...
+		color.Red("Error: %v", err)
+		os.Exit(1)
+	}
+	fmt.Fprintf(w, string(output))
+
+}
+
+func (s *K8sGPTServer) Serve() error {
+	http.HandleFunc("/analyze", s.analyzeHandler)
+	color.Green("Starting server on port " + s.Port)
```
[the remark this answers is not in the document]
thschue: fixed

pkg/server/main.go
thschue: +1 from my side and fixed

AlexsJones (approved):

PR #269, opened 2023-04-13
feat: running local models

### LINKED ISSUE #188: Local  model support
- LLaMa
- GPT4ALL

mudler:
~~This is an untested draft still~~ It works! see my comment below. This should be enough for testing on an OpenAI compatible endpoint just by letting the user change the base_url, so should work as well with https://github.com/go-skynet/LocalAI

I'll experiment a bit locally and refine this, adding docs too.  cc: @arbreezy @AlexsJones 

Closes #188 

## ✅ Checks


mudler:
works beautifully :)

I deployed something with a wrong tag, so it was failing to pull the image and bring up the service, that was the response (it took some time, but hey, that's now all local!):

    
    base ❯ ./k8sgpt analyze --explain                                                                                                                                                             
     100% |██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| (2/2, 1 it/min)            
                                                                                                                                                                                                                       
    0 llama/llama-69b7785db9-rlv9g(Deployment/llama)                                                                                                                                              
    - Error: Back-off pulling image "quay.io/go-skynet/llama-cli:fff"                                                                                                                             
    ```The message is: Back-off pulling image "quay.io/go-skynet/llama-cli:fff"                                                                                                                                        
                                                                                                                                                                                                  
    The solution is:                                                                                                                                                                              
                                                                                                                                                                                                  
    The error message indicates that there is a conflict in the image name being used in the Kubernetes deployment. To resolve this issue, you can either rename the image or update the image tag in the Kubernetes deployment file.
                                                                                                                                                                                                  
    For example, if the image name is "llama-cli:fff", you can rename it to something like "llama-cli:tag-1". Then, update the image tag in the Kubernetes deployment file to reflect the new name.                                              
                                                                                                                                                                                                  
    Alternatively, if the image name is not a critical part of the deployment, you can rename the image without affecting the deployment.                                                         
                                                                                                                                                                                                  
    1 llama/llama(llama)                                                                                                                                                                          
    - Error: Service has not ready endpoints, pods: [Pod/llama-69b7785db9-rlv9g], expected 1                                                                                                      
                                                                                                                                                                                                                       
    ```                                                                                                                                                                                           
    Service has not ready endpoints, pods: [Pod/llama-69b7785db9-rlv9g], expected 1                                                                                                               
                                                   
    ```                                                 
                                                                                                                                                                                                  
    The error message is indicating that the service does not have any available endpoints, and there is only one pod that is running. This means that there is no endpoint available to handle the traffic.                                     
                                                                                                                                                                                                                       
    To resolve this issue, you can either:                                                                                                                                                        
    
    1.   Deploy a new pod to the service, or
    2.   Update the endpoints of the existing pod to include the new endpoint.                                                                                                                                         
    
    If you are using the first option, you can use the kubectl apply command to deploy a new pod to the service. For example:                                                                                                                    
                                                                                                   
                                                                                                                                                                                                                       
    ```                                            
    kubectl apply -f https://raw.githubusercontent.com/Homebrew/kubeadm/main/examples/deploy-pod.yaml                                                                                                                                            
                                                                                                                                                                                                  
    ```                                                                                                                                                                                                                
    
    If you are using the second option, you can use the kubectl apply command to update the endpoints of the existing pod. For example:                                                                                                          
    
                                                                                                             
    ```                                                                                            
    kubectl apply -f https://raw.githubusercontent.com/Homebrew/kubeadm/main/examples/update-endpoints.yaml                                                                                                                                      
                                                        
    ```                                                 
    
    Note that you need to replace the values in the examples file with your own values.                                   
    
My setup:

```
base ❯ cat ~/.k8sgpt.yaml
ai:
    providers:
        - base_url: http://localhost:8080/v1
          model: ggml-koala-7b-model-q4_0-r2
          name: openai
kubeconfig: ....
kubecontext: ""
```

To bring up the server: 
- Follow the steps in and start the API: https://github.com/go-skynet/llama-cli#usage (I've used the koala model)
- Add the default koala prompt template for the model:
- ggml-koala-7b-model-q4_0-r2.bin.tmpl :
```
BEGINNING OF CONVERSATION: USER: {{.Input}} GPT:
```

Will polish, and add docs alongside with the PR, stay tuned!

mudler:
here looks good! - let me know if that's the good direction @AlexsJones , I've also updated the docs, feedback is welcome :)

README.md
```diff
@@ -267,6 +267,58 @@ The Kubernetes system is trying to scale a StatefulSet named fake-deployment usi
 
 </details>
 
+## Running local models
+
+To run local models, it is possible to use OpenAI compatible APIs, for instance [llama-cli](https://github.com/go-skynet/llama-cli) which uses [llama.cpp](https://github.com/ggerganov/llama.cpp) to run inference on consumer-grade hardware. Models supported by `llama.cpp` for instance are Vicuna, Alpaca, LLaMA, and koala. 
+
+To run local inference, you need to downlooad the models first, for instance you can find `ggml` compatible models in [huggingface.com](https://huggingface.co/models?search=ggml).
+
+### Start the API server
```
AlexsJones, line 276: This is awesome, but can we put it in the <details> tag?

pkg/ai/openai.go
```diff
@@ -27,8 +28,12 @@ type OpenAIClient struct {
 	model    string
 }
 
-func (c *OpenAIClient) Configure(token string, model string, language string) error {
-	client := openai.NewClient(token)
+func (c *OpenAIClient) Configure(token string, model string, language string, baseURL string) error {
```
AlexsJones, line 31: It feels like we could give llama its own struct rather than modifying the OpenAI one, what do you think?

AlexsJones:
Looks very exciting @mudler , @matthisholleville @arbreezy perhaps we use this as another reason to have a configuration Object for AI rather than passing more strings. Also @mudler I think having your own AI struct might be easier ( yes its a little more work, but it eventually means less manipulation of the openai.go)

arbreezy:
This adds great flexibility to k8sgpt !
Just a small comment @mudler.
Since there is no authentication happening with llama we need to tweak the auth [cmd](https://github.com/k8sgpt-ai/k8sgpt/blob/main/cmd/auth/auth.go#L60), skipping the password/token/api-key user prompt if llama provider is enabled in k8sgpt.

pkg/ai/openai.go
mudler: I'm definitely ok with it, however it will look like mostly the same - behind the scenes it's going still to use the OpenAI client. I'll wrap my changes in a separate struct so we can compare and pick the one we like!

mudler:
Thanks for the feedback :) Going to work on your feedback into separate commits and be back at you as soon as possible!

mudler:
@arbreezy @AlexsJones should I keep the base_url option to the OpenAI backend? Or shall I keep it untouched? I'm fine both ways

README.md
mudler: Sure thing :+1:

AlexsJones:
> @arbreezy @AlexsJones do you want me to keep the `base_url` option to the OpenAI backend? Or shall I leave it untouched? I'm fine both ways - I think might be useful e.g. when using proxies, but it's your call

I think we can leave it in there unused, as you say it might be useful eventually!

README.md
mudler: done!

pkg/ai/openai.go
mudler: done!

cmd/auth/auth.go
```diff
@@ -68,11 +69,20 @@ var AuthCmd = &cobra.Command{
 			password = strings.TrimSpace(string(bytePassword))
 		}
 
+		baseURL := ""
+		if backend == "llama" {
```
mudler, line 73: I'm not very comfortable to put backend-specific code in auth, however, in order to avoid this I think some refactor would be needed (for instance,  ask for backend-specific settings programmatically) to the interfaces. I didn't felt should be part of this PR, so I went with a simple approach. 

Let me know your feedback, I think we can generalize that easier later and keep this PR's scope just to the local backend

pkg/ai/noopai.go
```diff
@@ -16,7 +17,7 @@ type NoOpAIClient struct {
 	model    string
 }
 
-func (c *NoOpAIClient) Configure(token string, model string, language string) error {
+func (c *NoOpAIClient) Configure(token string, model string, language string, baseURL string) error {
```
mudler, line 20: I do agree this is growing, but again, I hold off from any change not in scope with this PR - I think a follow up for a generic config or move to functional style would be needed here.

mudler:
@AlexsJones @arbreezy implemented your suggestions - added some comments inline with the changes introduced

matthisholleville:
Very useful feature ! Good job @mudler !!

cmd/auth/auth.go
matthisholleville: I totally agree.

I think we should refactor this part of the code to make it more flexible to adding new backends, especially since @arbreezy is working on adding a new backend it seems.

cmd/auth/auth.go
```diff
@@ -68,11 +69,20 @@ var AuthCmd = &cobra.Command{
...
+			fmt.Printf("Enter %s API Base URL (e.g. `http://localhost:8080/v1`): ", backend)
+			scanner := bufio.NewScanner(os.Stdin)
```
matthisholleville, line 75: shoudn't we use the following to be more consistent ?

```go
bytePassword, err := term.ReadPassword(int(syscall.Stdin))
			if err != nil {
				color.Red("Error reading %s Key from stdin: %s", backend,
					err.Error())
				os.Exit(1)
			}
			password = strings.TrimSpace(string(bytePassword))
```

mudler: I was thinking to that indeed, but then the baseurl is not something should be kept confidential and be hidden to the user, so I went with a simpler approach. Happy to revisit this

cmd/auth/auth.go
AlexsJones: Are you okay holding off a little until that PR is in @mudler ? I don't mind either way.

mudler: I'm fine either ways, at the end is your call!  I'm just a little bit concerned this one can start to grow too much out of scope and lead to conflicts. Is there a branch somewhere already? I could rebase on top of it and start to play with it

mudler: afterthought maybe merging this in before might be actually helpful when refactoring providers - it gives a concrete example on what other providers might need to support. But I'm fine as well rebasing my work on top of the refactoring after that's being merged. Just cc'ing @arbreezy so he is aware of what's needed for llama here :arrow_up:

arbreezy: hello @mudler @AlexsJones  !
I 've added a draft [PR](https://github.com/k8sgpt-ai/k8sgpt/pull/297) with a first iteration of introducing a configuration interface and also adding llama and azure open ai.
I just wanted to get all changes in a single place for this work, eventually this draft PR will break in individual components (config interface, azure open ai, llama, etc)

cmd/auth/auth.go
arbreezy: My take is having the baseURL as another flag, a) it's not sensitive data and b) the rest of non sensitive information get passed via flags, I have implemented that in my draft PR to showcase it

cmd/auth/auth.go
mudler: hey @arbreezy ! thanks for taking a stab at it ! It looks good to me!

maybe I'd just split the llama changes from it, so can be introduced separately here, what do you think? otherwise it doesn't have documentation with the PR - or I can just add documentation here. I have no issues with it, just let me know!

cmd/auth/auth.go
mudler: @arbreezy looks great! I really like the approach :+1:

cmd/auth/auth.go
arbreezy: Yes, it's only fair to leave llama to you , just wanted to demonstrate how will look like, if everyone is happy I will close the draft PR and introduce two smaller PRs around config changes and azure open ai.

cmd/auth/auth.go
matthisholleville: make sense 👍

cmd/auth/auth.go
mudler: have also some updates on my side due to support of GPT4ALL-J, will update based off from your work!

arbreezy: awesome, I will find time tomorrow to create the first PR which introduces config interface (I won't include any llama related code)

mudler:
Rebased from main, docs updated, and gave it another shot locally!

> ~/_git/k8sgpt local_models*
> base ❯ ./k8sgpt auth --backend localai --model ggml-koala-7b-model-q4_0-r2.bin --baseurl http://localhost:8080/v1
> Using localai as backend AI provider
> Enter localai Key: New provider added
> key added
> 
> ~/_git/k8sgpt local_models*
> base ❯ ./k8sgpt analyze --explain --backend localai
>  100% |█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| (2/2, 2 it/min)           
> 
> 0 llama/llama(llama)
> - Error: Service has not ready endpoints, pods: [Pod/llama-69b7785db9-rlv9g], expected 1
>  The error message is saying that the Kubernetes service for the pod "llama-69b7785db9-rlv9g" has not yet been created. To fix this, you can either wait for the service to be created or manually create the service.
> 
> If you want to wait for the service to be created, you can check the status of the service in the Kubernetes cluster using the kubectl get pods command and check if the pod is in a running state. Once the pod is in a running state, the service will be created automatically.
> 
> If you want to manually create the service, you can use the kubectl create service command to create the service and specify the name, type, selector, ports, and other details of the service.
> 
> It's also important to note that the "expected 1" part of the error message refers to the number of endpoints that the service is expected to have. If the pod is not running or if there is a problem with the pod, it will not have any endpoints, which is why the service is not ready.
> 1 llama/llama-69b7785db9-rlv9g(Deployment/llama)
> - Error: Back-off pulling image "quay.io/go-skynet/llama-cli:fff"
>  The error message "Back-off pulling image "quay.io/go-skynet/llama-cli:fff" indicates that there was a failure to pull an image from a Docker registry. The back-off mechanism is used to avoid too many failed pulls and to avoid overloading the Docker registry.
> 
> To resolve the issue, you can try the following steps:
> 
> 1.   Check that the Docker image exists and is accessible in the Docker registry.
> 2.   Check that the Docker image has the correct tag and that it matches the image you are trying to pull.
> 3.   Try pulling the Docker image again after waiting for a few minutes.
> 4.   If the problem persists, you can try using a different Docker registry or Docker image.
> 
> If none of the above steps work, you can also try checking the Docker logs for more information about the error.

mudler:
@arbreezy just noticed that you were already setting the baseURL in https://github.com/k8sgpt-ai/k8sgpt/pull/309 :facepalm: 

I created https://github.com/k8sgpt-ai/k8sgpt/pull/310 in case we want to merge that separately so we don't conflict each other PRs, I'm fine either way!

pkg/ai/iai.go
```diff
@@ -49,7 +51,7 @@ type AIProvider struct {
 	Name     string `mapstructure:"name"`
 	Model    string `mapstructure:"model"`
 	Password string `mapstructure:"password"`
-	BaseURL  string `mapstructure:"base_url"`
+	BaseURL  string `mapstructure:"baseurl"`
```
mudler, line 54: sorry about this. I've messed up during cherry-pick and rebase on the other PR, now works as intended (`baseurl` in place of `base_url`)

mudler:
There was a small bug introduced in the #310 PRs, sorry about that, I think it was an oversight between cherrypicks and rebases here.

@AlexsJones @arbreezy @matthisholleville rebased, re-tested locally, good to go here! please let me know if anything is missing!

matthisholleville:
I'll test this tonight! Thank you for your contribution

matthisholleville:
Hi,

I just tested it and here's the error I encountered:

```bash
➜  k8sgpt git:(9b914fb) ./k8sgpt analyze --explain --backend localai --namespace k8sgpt --no-cache
Warning: Legacy config file at `/Users/matthisholleville/.k8sgpt.yaml` detected! This file will be ignored!
   0% |                                                                                                                                                                                       | (0/1, 0 it/hr) [0s:0s]
Error: failed while calling AI provider localai: error, json: cannot unmarshal string into Go struct field ErrorResponse.error of type map[string]json.RawMessage
```

Here's my setup:

- using the [chatgpt-openai/ggml-model-whisper](https://huggingface.co/chatgpt-openai/ggml-model-whisper/blob/main/ggml-model-gpt-2-345M.bin) model.
- no error log on LocalAI container
- I followed the README documentation from A to Z, nothing exotic.

mudler:
> Hi,
> 
> I just tested it and here's the error I encountered:
> 
> ```shell
> ➜  k8sgpt git:(9b914fb) ./k8sgpt analyze --explain --backend localai --namespace k8sgpt --no-cache
> Warning: Legacy config file at `/Users/matthisholleville/.k8sgpt.yaml` detected! This file will be ignored!
>    0% |                                                                                                                                                                                       | (0/1, 0 it/hr) [0s:0s]
> Error: failed while calling AI provider localai: error, json: cannot unmarshal string into Go struct field ErrorResponse.error of type map[string]json.RawMessage
> ```

Good catch! that was a miss from my side, I just traced it back, now the API returns OpenAI error types: https://github.com/go-skynet/LocalAI/pull/80

> 
> Here's my setup:
> 
>     * using the [chatgpt-openai/ggml-model-whisper](https://huggingface.co/chatgpt-openai/ggml-model-whisper/blob/main/ggml-model-gpt-2-345M.bin) model.

This model is for "whisper" - it is not for LLM/GPTs. I've updated the docs with an e2e example for gpt4all-j: https://github.com/go-skynet/LocalAI#example-use-gpt4all-j-model

> 
>     * no error log on LocalAI container

My bad, docs weren't aligned, you need to turn debug mode (by either setting the DEBUG env var to true, or specifying it in the CLI). I've updated the docs as well!

arbreezy (approved):
OK LGTM, 
it's working great with a local model
I suggest to merge the localai.

matthisholleville:
if @arbreezy tested and it's ok, it's good for me too. Thank you very much for you contribution @mudler !

matthisholleville (approved):
LGTM

panpan0000:
Hey, @mudler , Can you contribute a step by step doc guide to download a pre-trained model from somewhere , and use your feature ? 
very excited for this .

PR #272, opened 2023-04-14
feat: add node analyzer

### LINKED ISSUE #210: feature: analyzer for Nodes
Checklist:

* [ ] Yes
* [x] No

It would be helpful to also look at nodes for potential problems. Node conditions are a good indicator that something is wrong and can point to the root cause of why certain workloads are not being scheduled.

I would suggest this analyzer as core, enabled by default.

Users can detect and troubleshoot problems of their nodes

yeahservice:
Closes #210

## 📑 Description
Adds a node analyzer, that checks the condition on nodes and reports failures accordingly. 

Conditions are currently checked according to https://kubernetes.io/docs/concepts/architecture/nodes/#condition. But since cloud providers might introduce their own NodeConditionTypes and a new NodeConditionStatus might be added, the check takes this into account and at worst reports them as false positives.

## ✅ Checks

## ℹ Additional Information
```
./k8sgpt filters list
Active: 
> Node
Unused: 
> CronJob
> Pod
> Deployment
> PersistentVolumeClaim
> Service
> Ingress
> StatefulSet
> ReplicaSet
> NetworkPolicy
> HorizontalPodAutoScaler
> PodDisruptionBudget
```

```
./k8sgpt analyze --explain 

0 node1(node1)
- Error: node1 has condition of type MemoryPressure, reason NodeStatusUnknown: Kubelet stopped posting node status.
- Error: node1 has condition of type DiskPressure, reason NodeStatusUnknown: Kubelet stopped posting node status.
- Error: node1 has condition of type PIDPressure, reason NodeStatusUnknown: Kubelet stopped posting node status.
- Error: node1 has condition of type Ready, reason NodeStatusUnknown: Kubelet stopped posting node status.
The error message is indicating that the Kubernetes node is experiencing memory, disk, and PID pressure. This is due to the Kubelet not posting node status. The reason for this could be one of several factors, such as network issues or the Kubelet process crashing.

To solve this issue, you should start by checking the status of the Kubelet process on the affected node. You can do this using the command "systemctl status kubelet". If the process is not running, you can try starting it manually using "systemctl start kubelet". If the Kubelet process crashes frequently, you may need to investigate why this is happening and take steps to prevent it from happening in the future.

Additionally, you should check the network connectivity between the affected node and the Kubernetes master. This can often be a source of issues with node status updates. If there are any issues with network connectivity, you may need to troubleshoot those as well.

```


AlexsJones (approved):
Excellent work thank you!

thschue (approved):
lgtm

PR #273, opened 2023-04-14
feat: add server metrics

matthisholleville:
Closes # 

## 📑 Description

This commit add `metrics` path & the analyzer error metrics in the codebase. 

The changes have been made across all analyzers and include the addition of a new metric with label values for the analyzer's name, analyzed object's name, and namespace. 

The metric's value is set to the length of the analyzer objects failures.

## ✅ Checks

## ℹ Additional Information

```bash
$ curl http://localhost:8080/metrics
# HELP analyzer_errors Number of errors detected by analyzer
# TYPE analyzer_errors gauge
analyzer_errors{analyzer_name="HorizontalPodAutoscaler",namespace="k8sgpt",object_name="fake-hpa"} 1
analyzer_errors{analyzer_name="Ingress",namespace="k8sgpt",object_name="fake-ingress"} 3
...
```


pkg/analyzer/analyzer.go
```diff
@@ -7,6 +7,15 @@ import (
 	"github.com/fatih/color"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/integration"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	AnalyzerErrorsMetric = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
thschue, line 15: I think it would make more sense to make this a counter (promauto.NewCounterVec)? 
```suggestion
	AnalyzerErrorsMetric = promauto.NewGaugeVec(prometheus.GaugeOpts{
```

thschue (commented):
Good work, I think the Metric should be a Counter instead of a Gauge, but the rest look good!

matthisholleville:
Thank you for your feedback Thomas. From my understanding, the metrics can increase or decrease depending on the number of issues. The user can partially fix errors and the metrics will decrease accordingly.

On the other hand, the counter can only increase which is not our case.

What do you think?

pkg/analyzer/analyzer.go
matthisholleville: I'm not sure because this metric can decrease over time. What do you think ?

AlexsJones (approved):

pkg/analyzer/analyzer.go
AlexsJones: I agree that having a gauge means we can do things like `rate[5m](AnalyzerErrors)` etc

thschue:
> Thank you for your feedback @thschue. From my understanding, the metrics can increase or decrease depending on the number of issues. The user can partially fix errors and the metrics will decrease accordingly.
> 
> On the other hand, the counter can only increase which is not our case.
> 
> What do you think?

Sorry, had the wrong use-case in mind. The gauge is ok in this case!

thschue (approved):
lgtm

PR #290, opened 2023-04-17
fix: deployment/cronjob analyzers namespace filtering

matthisholleville:
Closes # 

## 📑 Description

These changes address an issue with the Cronjob and Deployment analyzers. The `namespace` flag was not being taken into account for these two analyzers, resulting in returning all deployments/cronjobs from all namespaces of a cluster even if the `--namespace` flag was specified.

I took the opportunity to add tests to the existing analyzers to prevent regressions in the future.

## ✅ Checks


thschue (approved):
lgtm

PR #300, opened 2023-04-20
chore: analyze Pod ReadinessProbe faliure

panpan0000:
Closes # 

## 📑 Description

To handle situation like 
```
# kubectl get po
NAME                        READY   STATUS    RESTARTS   AGE
xxxxx                          0/1     Running   0          46m
```

## ✅ Checks


thschue (approved):
lgtm

pkg/analyzer/pod.go
```diff
@@ -70,6 +70,16 @@ func (PodAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 						})
 					}
 				}
+			} else {
+				// when pod is Running but its ReadinessProbe fails
+				if containerStatus.Ready == false && pod.Status.Phase == "Running" {
+					// parse the event log and append details
+					evt, err := FetchLatestEvent(a.Context, a.Client, pod.Namespace, pod.Name)
+					if err != nil || evt == nil {
+						continue
```
matthisholleville, line 79: 👋 Thank you for your contribution. 

I don't understand this else. Shouldn't we add append the failures table as indicated in your comment?

Can you also add a test case for this change please ?

matthisholleville (changes requested):

pkg/analyzer/pod.go
panpan0000: when the ReadinessProbe fail, the `containerStatus.State.Waiting` will be nil.

```
 kubectl get po
NAME                        READY   STATUS    RESTARTS   AGE
my-nginx-7fb4dbcf47-4ch4w   0/1     Running   0          3h58m
```

```
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
....
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2023-04-20T03:00:21Z"
      message: 'containers with unready status: [nginx]'
      reason: ContainersNotReady
      status: "False"
      type: ContainersReady
  containerStatuses:
    - containerID: containerd://b474843f8cd6d2d6e91d679bd329f8ea48ac1a3db7d7bdeb80879a77e2bcd2d0
      image: daocloud.io/daocloud/nginx:latest
      imageID: daocloud.io/daocloud/nginx@sha256:89a42c3ba15f09a3fbe39856bddacdf9e94cd03df7403cad4fc105088e268fc9
      lastState: {}
      name: nginx
      ready: false
      restartCount: 0
      started: true
      state:
        running:
          startedAt: "2023-04-20T03:00:51Z"
    phase: Running
```

matthisholleville: Yes I understand the need but I don't understand why we say in the commentary that we have to append a new failure to the board but we don't do it.

```bash
					HERE -> // parse the event log and append details
					evt, err := FetchLatestEvent(a.Context, a.Client, pod.Namespace, pod.Name)
					if err != nil || evt == nil {
						continue
```

panpan0000: I missed this part. Thank you for your reminder. 
I just added unit-test case.

actually, I just ran go test , but no good idea how to build a test binary.

matthisholleville (approved):

PR #306, opened 2023-04-20
feat: modify analyze error handling

### LINKED ISSUE #289: Error: `the server could not find the requested resource`
Checklist:

### Subject of the issue

k8sgpt tries to read some resource that does not exist on cluster.

### Your environment

* Version of Kubernetes
  C: `v1.26.3`, S: `1.16.11`
* Host OS and its version / If windows, is it WSL? `macOS`
* Version of k8sgpt `0.2.3`

### Steps to reproduce

```
$ k8sgpt analyze --kubeconfig ~/.kube/my-cluster
Error: the server could not find the requested resource
```

### Expected behaviour
It should work? _Which_ requested resource is not found? It'd better UX if k8sgpt keep working on any error case. Is this resource really critical that we need to _exit_ immediately?

It'd better put the `api-resource` list in README that k8sgpt is using.

### Actual behaviour

```
Error: the server could not find the requested resource
```

matthisholleville:
Closes #289 

## 📑 Description

modify error handling to return a list of errors to display to the user at the end of analysis without blocking it if an error is detected (e.g., version of an object is not available on user's cluster)

## ✅ Checks

## ℹ Additional Information

```bash
(matthisholleville) ➜  k8sgpt git:(main) ✗ ./k8sgpt analyze --namespace k8sgpt
Error: CronJob the server could not find the requested resource

0 k8sgpt/fake-hpa(fake-hpa)
- Error: HorizontalPodAutoscaler uses StatefulSet/fake-deployment as ScaleTargetRef which does not exist.
```

### DIFF
--- a/cmd/analyze/analyze.go
+++ b/cmd/analyze/analyze.go
@@ -36,10 +36,11 @@ var AnalyzeCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		err = config.RunAnalysis()
-		if err != nil {
-			color.Red("Error: %v", err)
-			os.Exit(1)
+		analysisErrors := config.RunAnalysis()
+		if len(analysisErrors) != 0 {
+			for _, err := range analysisErrors {
+				color.Red("Error: %s", err)
+			}
 		}
 
 		if explain {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -55,12 +55,14 @@ func (s *Config) analyzeHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	err = config.RunAnalysis()
-	if err != nil {
-		color.Red("Error: %v", err)
+	analysisErrors := config.RunAnalysis()
+	if analysisErrors != nil {
+		var errorMessage string
+		for _, err := range analysisErrors {
+			errorMessage += err.Error() + "\n"
+		}
+		http.Error(w, errorMessage, http.StatusInternalServerError)
 		health.Failure++
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
 	}
 
 	if explain {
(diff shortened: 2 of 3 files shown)


cmd/analyze/analyze.go
```diff
@@ -36,10 +36,11 @@ var AnalyzeCmd = &cobra.Command{
...
 		}
 
-		err = config.RunAnalysis()
-		if err != nil {
-			color.Red("Error: %v", err)
-			os.Exit(1)
+		analysisErrors := config.RunAnalysis()
+		if len(analysisErrors) != 0 {
+			for _, err := range analysisErrors {
+				color.White("Error: %s", err)
```
AlexsJones, line 42: We might want another colour as that won't show on default terminals

pkg/server/server.go
```diff
@@ -55,12 +55,12 @@ func (s *Config) analyzeHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	err = config.RunAnalysis()
-	if err != nil {
-		color.Red("Error: %v", err)
+	analysisErrors := config.RunAnalysis()
+	if analysisErrors != nil {
+		for _, err := range analysisErrors {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
AlexsJones, line 61: Won't the first http Error return the function?

matthisholleville: Bien vu ! Je vais modifier cela

cmd/analyze/analyze.go
matthisholleville: The Kubernetes client returns this kind of warning if we try to parse a deprecated APIVersion. I chose this color to maintain the same style, but I can change it
Screenshot 2023-04-20 at 14 10 27

PR #308, opened 2023-04-20
feat: add configuration interface to support customer providers

arbreezy:
Closes # 

## 📑 Description
Refactoring the way we configure k8sgpt ai providers by adding an interface.

## ✅ Checks

## ℹ Additional Information

This PR is a prerequisite for adding azure openai and llama providers.

### DIFF
--- a/pkg/ai/iai.go
+++ b/pkg/ai/iai.go
@@ -5,10 +5,15 @@ import (
 )
 
 type IAI interface {
-	Configure(token string, model string, language string) error
+	Configure(config IAIConfig, language string) error
 	GetCompletion(ctx context.Context, prompt string) (string, error)
 	Parse(ctx context.Context, prompt []string, nocache bool) (string, error)
 	GetName() string
+}
+
+type IAIConfig interface {
+	GetPassword() string
+	GetModel() string
 }
 
 func NewClient(provider string) IAI {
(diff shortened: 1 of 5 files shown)


pkg/ai/iai.go
```diff
@@ -5,12 +5,17 @@ import (
...
 type IAI interface {
-	Configure(token string, model string, language string) error
+	Configure(config IAIConfig, language string) error
 	GetCompletion(ctx context.Context, prompt string) (string, error)
 	Parse(ctx context.Context, prompt []string, nocache bool) (string, error)
 	GetName() string
 }
 
+type IAIConfig interface {
+	GetToken() string
```
matthisholleville, line 15: To be more generic, shouldn't rename this to `GetPassword` ?

pkg/ai/prompts.go
```diff
@@ -0,0 +1,8 @@
+package ai
```
matthisholleville, line 1: This is just a suggestion but shouldn't we rename this file to `common.go` ?

matthisholleville (commented):
Good job again @arbreezy !

pkg/ai/prompts.go
arbreezy: I think `prompts.go` is more intuitive tbh if we see a need for more common functionality we rename then

pkg/ai/iai.go
arbreezy: fair !

matthisholleville (approved):

Issue #117, opened 2023-03-28
Allow for support for Azure OpenAI Services


msandbu:
Since many (including myself) with organizations will be using ChatGPT as part of Azure OpenAI Service. The feature should also allow the ability to point to ChatGPT running on Azure OpenAI Services. 

It is using the same API but different endpoint https://openaichatgpt.openai.azure.com, could for instance be a featureflag to point it to another endpoint?

AlexsJones:
We have the ability to switch the `--backend` and an `IAI` interface which should make implementing it pretty easily. 
Do you know if there is a golang based SDK? The current SDK we use is tightly coupled to OpenAI public endpoints though I could look at whether we can modify that.

msandbu:
Are you using this? https://github.com/sashabaranov/go-openai (Do not believe there is
Perhaps it would be better if I get support for the Azure OpenAI endpoint there first make it easier to add the new backend here?

roberthstrand:
I'm looking into this going forward. The SDK we are using has an open issue on support for Azure OpenAI, but I will look into several solutions for this.

AlexsJones:
I feel we might want to build our own SDK to control our destiny a bit more.

roberthstrand:
I was thinking of that as well. Let's have a discussion about it tomorrow, when I got more time :)

stulzq:
This is a solution：https://github.com/stulzq/azure-openai-proxy

jnytnai0613:
> Are you using this? https://github.com/sashabaranov/go-openai (Do not believe there is Perhaps it would be better if I get support for the Azure OpenAI endpoint there first make it easier to add the new backend here?

Hi!
It seems that [sashabaranov/go-openai](https://github.com/sashabaranov/go-openai/pull/214) already supports Azure OpenAI.
https://github.com/sashabaranov/go-openai/releases/tag/v1.6.0

arbreezy:
I have created a draft PR to introduce azure open ai provider, bear with us a little longer as it also includes other changes that need to test first before merging it

arbreezy:
@msandbu , would you be able to beta-test the feature since I don't have an azure openai deployment up and running?

PR #309, opened 2023-04-20
feat: add azure openai provider

arbreezy:
Closes https://github.com/k8sgpt-ai/k8sgpt/issues/117

## 📑 Description
This PR introduces azure openai and get viper to omit empty strings when writing to config files

## ✅ Checks


matthisholleville:
Hello @arbreezy , shouldn't we update the README to indicate that we have a new AI Backend available?

arbreezy:
> Hello @arbreezy , shouldn't we update the README to indicate that we have a new AI Backend available?

Yes we will add docs, I just didn't have much time yesterday; I want to test it as well before merging.

roberthstrand:
Tried it out, and my first observation is that the help text needs to updated to reflect these changes. I had to check the code to find out what to define as the backend.

Secondly, and this might be out of scope for the actual implementation, but if I want to do an analysis with the Azure OpenAI backend after I have authenticated, I would have so explicitly state that I'm using that as a backend, even when I have no other backends in my configuration file. @AlexsJones, this is probably something we can look into in another issue, right?

Third, and more importantly, I get an error when I run analyze with explain. This is the commands I ran:

```
> go run . auth -u https://rsopenaigptdemo.openai.azure.com/ -e rsk8sgpt -b azureopenai
Using azureopenai as backend AI provider
Enter azureopenai Key: Provider updated
key added

> go run . analyse -b azureopenai -e
   0% |                                                                                                            | (0/3, 0 it/hr) [0s:0s]
Error: failed while calling AI provider azureopenai: Post "/openai/deployments//chat/completions?api-version=2023-03-15-preview": unsupported protocol scheme ""
exit status 1
```

arbreezy:
> Tried it out, and my first observation is that the help text needs to updated to reflect these changes. I had to check the code to find out what to define as the backend.
> 
> Secondly, and this might be out of scope for the actual implementation, but if I want to do an analysis with the Azure OpenAI backend after I have authenticated, I would have so explicitly state that I'm using that as a backend, even when I have no other backends in my configuration file. @AlexsJones, this is probably something we can look into in another issue, right?
> 
> Third, and more importantly, I get an error when I run analyze with explain. This is the commands I ran:
> 
> ```
> > go run . auth -u https://rsopenaigptdemo.openai.azure.com/ -e rsk8sgpt -b azureopenai
> Using azureopenai as backend AI provider
> Enter azureopenai Key: Provider updated
> key added
> 
> > go run . analyse -b azureopenai -e
>    0% |                                                                                                            | (0/3, 0 it/hr) [0s:0s]
> Error: failed while calling AI provider azureopenai: Post "/openai/deployments//chat/completions?api-version=2023-03-15-preview": unsupported protocol scheme ""
> exit status 1
> ```

hello @roberthstrand,
Documentation has been added along with some basic checks when running `k8sgpt auth` to ensure you will use a valid ai provider/backend.
Could you please run again a test with your Azure OpenAI deployment ?

Re: the analyze cli , I will address it, in a different PR; I agree there is no need to specify a backend if it's the only one in your configuration

Thanks in advance :) !

AlexsJones:
ready for review?

arbreezy:
> ready for review?

yes !

pkg/util/util.go
```diff
@@ -185,3 +185,17 @@ func EnsureDirExists(dir string) error {
 
 	return err
 }
+
+func Backends() []string {
```
AlexsJones, line 189: I would prefer not to have this in the utils file as its another place to update. Can we have a function in the iai file?

pkg/ai/iai.go
```diff
@@ -30,12 +30,15 @@ type IAIConfig interface {
 	GetPassword() string
 	GetModel() string
 	GetBaseURL() string
+	GetEngine() string
 }
```
AlexsJones, line 35: I would suggest doing something like 
```
var (
backends := []IAI { 
 &OpenAIClient{}, 
&AzureAIClient{}, 
...
}
```

Then in the new client, create a loop 

```
for b, := backend {
 // if it matches b.GetName() return
}

````

File: pkg/ai/iai.go [hunk 3]
```diff
@@ -52,8 +55,9 @@ type AIConfiguration struct {
 type AIProvider struct {
 	Name     string `mapstructure:"name"`
 	Model    string `mapstructure:"model"`
-	Password string `mapstructure:"password"`
-	BaseURL  string `mapstructure:"baseurl"`
+	Password string `mapstructure:"password" yaml:"password,omitempty"`
+	BaseURL  string `mapstructure:"baseurl" yaml:"baseurl,omitempty"`
+	Engine   string `mapstructure:"engine" yaml:"engine,omitempty"`
```
AlexsJones, line 60: What is engine?

pkg/util/util.go
arbreezy: I will remove the function and have it as part of vars in iai file along with the ai clients.

pkg/ai/iai.go
AlexsJones: Can't we just map `Model` to this?

arbreezy: I believe not, model is still needed in chat completions' request.

pkg/ai/iai.go
arbreezy: I updated the PR, based on your feedback

pkg/ai/iai.go
AlexsJones: Perhaps it would make sense in this very custom case to have a unique struct that uses the IAI interface?

arbreezy: I get your point and tbf I thought it was a small addition/change in order to need a new struct whilst working with meta tags there's no trace of engine in user's config file unless it's set by them.
Said that, if it's not readable I can follow up with another implementation of the interface.

cmd/auth/auth.go
```diff
@@ -117,4 +120,6 @@ func init() {
 	AuthCmd.Flags().StringVarP(&password, "password", "p", "", "Backend AI password")
 	// add flag for url
 	AuthCmd.Flags().StringVarP(&baseURL, "baseurl", "u", "", "URL AI provider, (e.g `http://localhost:8080/v1`)")
+	// add flag for azure open ai engine/deployment name
+	AuthCmd.Flags().StringVarP(&engine, "engine", "e", "", "Azure AI deployment name")
```
AlexsJones, line 124: This needs to be a `MarkFlagRequired` right? It's not optional for this command

cmd/serve/serve.go
```diff
@@ -48,11 +48,10 @@ var ServeCmd = &cobra.Command{
...
 			model := os.Getenv("K8SGPT_MODEL")
 			baseURL := os.Getenv("K8SGPT_BASEURL")
-
-			// If the envs are set, allocate in place to the aiProvider
+			engine := os.Getenv("K8SGPT_ENGINE")
+			// If the envs are set, alocate in place to the aiProvider
 			// else exit with error
-			envIsSet := backend != "" || password != "" || model != "" || baseURL != ""
-
+			envIsSet := backend != "" || password != "" || model != "" || baseURL != "" || engine != ""
```
AlexsJones, line 54: Engine should not be mandatory for serve under different backends

pkg/util/util.go
```diff
@@ -185,3 +185,13 @@ func EnsureDirExists(dir string) error {
 
 	return err
 }
+
+func ValidBackend(validBackends []string, backend string) bool {
```
AlexsJones, line 189: does this really need to exist here? Isn't this a generic find in slice?

AlexsJones (changes requested):

cmd/serve/serve.go
arbreezy: I think same goes for baseurl.
OpenAI's backend  doesn't require a base url to work.
WDYT?

cmd/auth/auth.go
arbreezy: thanks !
I've marked both engine and baseurl if backend is `azureopenai`

cmd/serve/serve.go
AlexsJones: Agreed

pkg/util/util.go
arbreezy: I will resolve the comment but please re-open if you don't like the implementation

README.md
```diff
@@ -258,9 +258,33 @@ _Analysis with serve mode_
...
+Lastly, enter your Azure API key, after the prompt.
+
+Now you are ready to analyze with the azure openai backend:  
+```
+k8sgpt analyze --explain --backend azureopenai
+```
 
 </details>
 
+### Running local models
```
AlexsJones, line 285: This merge looks wrong, can you shift down the anonymisation header below the AI/models stuff?

cmd/auth/auth.go
```diff
@@ -31,13 +31,21 @@ var (
...
 )
 
 // authCmd represents the auth command
 var AuthCmd = &cobra.Command{
 	Use:   "auth",
 	Short: "Authenticate with your chosen backend",
 	Long:  `Provide the necessary credentials to authenticate with your chosen backend.`,
+	PreRun: func(cmd *cobra.Command, args []string) {
+		backend, _ := cmd.Flags().GetString("backend")
+		if backend == "azureopenai" {
```
AlexsJones, line 44: This might sound pedantic, but I would advise `strings.ToLower(backend)` on that evaluation

cmd/serve/serve.go
```diff
@@ -47,17 +47,17 @@ var ServeCmd = &cobra.Command{
 			password := os.Getenv("K8SGPT_PASSWORD")
 			model := os.Getenv("K8SGPT_MODEL")
 			baseURL := os.Getenv("K8SGPT_BASEURL")
-
+			engine := os.Getenv("K8SGPT_ENGINE")
```
AlexsJones, line 50: Just a reminder but this will need to be added into the operator.

pkg/ai/azureopenai.go
```diff
@@ -0,0 +1,98 @@
...
+	return resp.Choices[0].Message.Content, nil
+}
+
+func (a *AzureAIClient) Parse(ctx context.Context, prompt []string, cache cache.ICache) (string, error) {
+	inputKey := strings.Join(prompt, " ")
+	// Check for cached data
+	sEnc := base64.StdEncoding.EncodeToString([]byte(inputKey))
+	cacheKey := util.GetCacheKey(a.GetName(), a.language, sEnc)
+	// find in viper cache
+	if cache.Exists(cacheKey) {
```
AlexsJones, line 62: This looks to be missing `!cache.IsCacheDisabled()`  which is the `--no-cache=` flag

AlexsJones (changes requested):

AlexsJones (dismissed):

README.md
arbreezy: sorted !

AlexsJones (approved):

shubham-uipath:
is there any documentation change required for this for end users ? wanted to validate this once with my azure openai account .

PR #311, opened 2023-04-21
feat: async calls

AlexsJones:
Closes # 

## ✅ Checks


matthisholleville:
Great job, Alex! It looks like the client is experiencing a bit of throttling on one of my clusters with 100 namespaces and 500 deployments:

```bash
I0421 09:04:25.405611 32626 request.go:690] Waited for 1.128472333s due to client-side throttling, not priority and fairness, request: GET:...
```

To avoid this warning and prevent overloading the Kubernetes server, should we consider using channels to synchronize our Go routines? This will help us wait for the completion of each request before sending another.

AlexsJones:
> To avoid this warning and prevent overloading the Kubernetes server, should we consider using channels to synchronize our Go routines? This will help us wait for the completion of each request before sending another.

Can you give me an example of what you're thinking? Sounds like using channels to make it synchronous again. 

We could batch the routines or to add slight staggering of the requests also

AlexsJones:
Ill take a look using a buffered channel

pkg/analysis/analysis.go
```diff
@@ -35,6 +35,7 @@ type AnalysisStatus string
 const (
 	StateOK              AnalysisStatus = "OK"
 	StateProblemDetected AnalysisStatus = "ProblemDetected"
+	MaxConcurrent        int            = 10
```
matthisholleville, line 38: Nice! Shouldn't we add a flag `--concurrency` to the analyzer so that we can set the maximum concurrent value?

AlexsJones: Yep leave it with me

matthisholleville (approved):

PR #324, opened 2023-04-24
feat: use OS conform path for storing cached results

### LINKED ISSUE #323: feature: Store cache in XDG specified cache directory
Checklist:

* [ ] Yes
* [x] No

As with the config location (solved in #235) the XDG specification specifies where to store 
cache files. So instead of caching results in the user config, there should be individual files created
in the corresponding caching location of the OS (on linux this would be `~/.cache`). 

The user config could be treated as immutable and cached files are stored in the corresponding folder the OS
specifies. 

## Potential drawbacks

I could not think of any drawback.

patrickpichler:
Closes #323

## 📑 Description

Instead of storing cached values in the config yaml, they are now stored under these OS specific locations:
* Linux: `~/.cache/k8sgpt`
* MacOS: `~/Library/Caches`
* Windows: `%LocalAppData%\cache`

Additionally a `Cache` package and interface has been introduced. Currently there are two implementations:
* Noop - Doesn't do anything
* FileBased - Stores data in files under the locations listed above

## ✅ Checks

- [] My code requires changes to the documentation

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1,0 +1,15 @@
+package cache
+
+type ICache interface {
+	Store(key string, data string) error
+	Load(key string) (string, error)
+	Exists(key string) bool
+}
+
+func New(noCache bool) ICache {
+	if noCache {
+		return &NoopCache{}
+	}
+
+	return &FileBasedCache{}
+}
--- a/pkg/ai/iai.go
+++ b/pkg/ai/iai.go
@@ -15,12 +15,14 @@ package ai
 
 import (
 	"context"
+
+	"github.com/k8sgpt-ai/k8sgpt/pkg/cache"
 )
 
 type IAI interface {
 	Configure(config IAIConfig, language string) error
 	GetCompletion(ctx context.Context, prompt string) (string, error)
-	Parse(ctx context.Context, prompt []string, nocache bool) (string, error)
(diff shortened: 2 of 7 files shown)


pkg/cache/cache.go
```diff
@@ -0,0 +1,15 @@
+package cache
+
+type Cache interface {
```
matthisholleville, line 3: Wouldn't it be more consistent with the rest of the code to rename this to `ICache`?

matthisholleville (commented):
Thank you for your great contribution again ! As mentioned on Slack, have you run any tests on the Docker image?

pkg/cache/cache.go
patrickpichler: Good point! I adjusted it :slightly_smiling_face:

patrickpichler:
@matthisholleville I did some quick tests in docker images and it should work as expected, since
the distroless static container sets and creates a home directory :slightly_smiling_face:

thschue (approved):
lgtm

pkg/cache/cache.go
patrickpichler: @matthisholleville what is the workflow here? 
Should I resolve the conversation, or are you going to do it (I would assume you click the button, since you requested the change, but I might be wrong :slightly_smiling_face: )?

PR #326, opened 2023-04-25
fix: explicitly pass in filter to async analysis go routine

patrickpichler:
## 📑 Description

Before the filter inside the func literal was capturing the value from the outer loop. This is a subtle mistake, since in combination with running the function literal as go routine, the value of filter could have already changed at invocation time.

To fix this, the filter is now passed in as an argument to the func literal.

## ✅ Checks


matthisholleville (approved):
lgtm

PR #368, opened 2023-05-02
feat: rework output format

### LINKED ISSUE #325: question: -o json is not returning an entire json structure for some cases
Checklist:

### Subject of the issue

Run the comand `./k8sgpt analyze --explain -o json` is returning json and text (not just json)

### Your environment

Local macbook M2, MacOS, docker, gcp cluster

* Version of Kubernetes
  v1.26.1
* Host OS and its version / If windows, is it WSL?
 Mac OS
* Version of k8sgpt
 v0.2.4

### Steps to reproduce

* Step 1
Create a cluster in gcp
* Step 2
Run analyze using `-o json` option

### Expected behaviour
The system should return a json estructure

### Actual behaviour

the system is returning, text lines out of json broking the json dumps

### Additional context / screenshots
The problem is here 

https://github.com/k8sgpt-ai/k8sgpt/blob/c3cc413e7fc3b06b310779dfa3cb4863ea9f3ed2/pkg/analyzer/service.go#L50

The system is returning the line with the text instead add in the failure array together other failures, the byte stream returned both text and the json so is impossible to convert in a right key value structure

matthisholleville:
Closes #325 

## 📑 Description

I have revamped the error handling to ensure that they are included and adhere to the output format. Now we have an `errors` attribute in the `Analysis` object that collects all the errors and displays them based on the output selected by the user.

I also deleted some not useful logs on configuration management

## ✅ Checks

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -39,7 +39,7 @@ type Analysis struct {
 	Client         *kubernetes.Client
 	AIClient       ai.IAI
 	Results        []common.Result
-	Errors         []string
+	Errors         []error
 	Namespace      string
 	Cache          cache.ICache
 	Explain        bool
@@ -116,7 +116,7 @@ func NewAnalysis(backend string, language string, filters []string, namespace st
 	}, nil
 }
 
-func (a *Analysis) RunAnalysis() {
+func (a *Analysis) RunAnalysis() []error {
 	activeFilters := viper.GetStringSlice("active_filters")
 
 	analyzerMap := analyzer.GetAnalyzerMap()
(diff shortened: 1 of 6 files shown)


pkg/analysis/analysis.go
```diff
@@ -175,11 +176,10 @@ func (a *Analysis) RunAnalysis() []error {
 					<-semaphore
 				}(analyzer, filter)
 			} else {
-				errorList = append(errorList, fmt.Errorf(fmt.Sprintf("\"%s\" filter does not exist. Please run k8sgpt filters list.", filter)))
+				a.Errors = append(a.Errors, fmt.Sprintf(fmt.Sprintf("\"%s\" filter does not exist. Please run k8sgpt filters list.", filter)))
```
AlexsJones, line 179: It occurs to me this needs a mutex around it

matthisholleville: I think, it is not necessary to use mutex.Lock() and mutex.Unlock() around the else statement. Since there is no concurrent access or modification of shared variables in the else block, you don't need to use a mutex lock there. What do you think ?

AlexsJones (approved):

PR #369, opened 2023-05-02
feat: add auth commands

HatiCode:
## 📑 Description
Initiated a change on the auth command the follow the style of "filter"
This allows for now to: 
* list the configured providers
* Add a new provider configuration

Still needed (If PR is accepted an issue for this will be created):
* possibility to update a previously configured provider

Older Flags are now defined in the "new" command.
Creating a provider would look like:

`k8sgpt auth new -b -m -p -u`

## ✅ Checks


AlexsJones:
It feels quite nice to separate these out, I would ask you update the documentation in the README and the docs page though

AlexsJones (commented):
Please update the README

cmd/auth/list.go
```diff
@@ -0,0 +1,47 @@
...
+
+		// get ai configuration
+		err := viper.UnmarshalKey("ai", &configAI)
+		if err != nil {
+			color.Red("Error: %v", err)
+			os.Exit(1)
+		}
+
+		// iterate over the provider list and prints each provider name
+		for _, provider := range configAI.Providers {
```
AlexsJones, line 39: @matthisholleville does this look right or will it just show the currently set provider?

AlexsJones: I think it would be a really interesting thing to implement, we would need to make sure that if they select the new provider and it doesn't have credentials it goes through the auth flow again

AlexsJones (approved):

PR #394, opened 2023-05-10
feat: improve `analyze` command

### LINKED ISSUE #332: Some tweaks for the terminal output for readability
Checklist:
* [X] I've discussed this feature in the #k8sgpt slack channel

* [ ] Yes
* [X] No

Current output of `analyze` command is slightly confusing and hard to read. Let's discuss more about how can we make this more clean.

Screenshot 2023-04-25 at 12 28 20

Some ideas:

* Hide `the server could not find the requested resource` errors by default (enable in verbose or details mode)
* Hide internal cache related errors
* Add `\n` between yellow sections and top of yellow GPT response (for more readable)
* Index should start from `1` (instead `0`)
* Omit parentheses if value nil or empty (for example, `()`)

Clean and better output.

## Potential drawbacks
No drawbacks.

rakshitgondwal:
Closes #332

## 📑 Description
Made the following changes-
- Added `\n` between yellow sections and the top of the yellow GPT response for better readability.
- Made the index start from 1 instead 0 for the results.
- Added index for the errors too.
- Parentheses would be omitted if the value of `result.ParentObject` would be empty.

Here's a reference image for these issues- 
234235472-8dc723bd-1dea-4666-b99a-630605c54be4

## ✅ Checks

### DIFF
--- a/pkg/analysis/output.go
+++ b/pkg/analysis/output.go
@@ -58,9 +58,9 @@ func (a *Analysis) textOutput() ([]byte, error) {
 	var output strings.Builder
 	if len(a.Errors) != 0 {
 		output.WriteString("\n")
-		output.WriteString(color.YellowString("Warnings : \n"))
-		for _, aerror := range a.Errors {
-			output.WriteString(fmt.Sprintf("- %s\n", color.YellowString(aerror)))
+		output.WriteString(color.YellowString("Warnings:\n"))
+		for i, aerror := range a.Errors {
+			output.WriteString(fmt.Sprintf("%d. %s\n", i+1, color.YellowString(aerror)))
 		}
 	}
 	output.WriteString("\n")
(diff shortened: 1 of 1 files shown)


rakshitgondwal:
There were two more changes suggested-
- Hide the `server could not find the requested resource` errors by default (enable in verbose or details mode)
- Hide internal cache related errors

but I am not being able to get around these. Can I get some help here?

matthisholleville:
I pushed a commit quickly, I think it can inspire you for the rework of the logs concerning the cache : https://github.com/k8sgpt-ai/k8sgpt/compare/main...matthisholleville:k8sgpt:feat/rework-cache-log

AlexsJones:
> Made the index start from 1 instead 0.

👀

rakshitgondwal:
> > Made the index start from 1 instead 0.
> 
> eyes

For the results*

rakshitgondwal:
Hey @matthisholleville 
From what I can understand, after going through your commit, we are now handling the cache errors in the `ai` package itself, rather than handling it in `file_based.go` but how can we hide the internal system errors?
I was thinking of just returning `false` from the `file_based.go` but that won't be a good idea, or would it be?

matthisholleville:
I think we should add a global argument to handle the verbosity of these logs

rakshitgondwal:
> I think we should add a global argument to handle the verbosity of these logs

Ohh something like if a user provides `--versbose` then we print the logs else we just return something like `error while testing if cache key exists, use --verbose to get more details`

AlexsJones:
Is this ready for review?

rakshitgondwal:
> Is this ready for review?

Nope, still have this question that do we introduce a `--verbose` flag to print the logs? I think Matthis is busy so can you help me out here?

PR #403, opened 2023-05-12
chore: customized prompt template for integration plugins

panpan0000:
Closes # NA 

## 📑 Description

When there are more integration plugin, the default prompt may not be enough.
So this PR provides the flexibility to customize prompt template for different resource kind (mainly for those introduced by  integration plugin )  and provide an example for trivy.

## ✅ Checks


pkg/ai/prompts.go
```diff
@@ -2,4 +2,10 @@ package ai
 
 const (
 	default_prompt = "Simplify the following Kubernetes error message and provide a solution in %s: %s"
+	trivy_prompt   = "Explain the following trivy scan result and the detail risk or root cause of the CVE ID, then provide a solution. DO NOT explain if you cannot access attached URLs. Response in %s: %s"
```
AlexsJones, line 5: What is the thinking behind the aquasec URLs? Those are quite useful for an engineer to have context around the CVE from the NVD

panpan0000: Hi, @AlexsJones , Thanks for your review.
Last time , chatGPT will complain that it cannot access the attached NVD URL page for detail. So I tried to suppress its  `sorry I cannot access xxx` wording. But today seems it gets smarter when I tried again.
I will remove that line . thanks

panpan0000: all set now

AlexsJones (approved):

PR #406, opened 2023-05-14
Improve default prompt

camigira:
## 📑 Description
Attempt to improve the `default_prompt` aiming to:
- Prevent Prompt Injection by specifying variable text win triple dashes `-`.
- Keep output concise but insightful.
- Clearer output Error vs Solution.

**Original Prompt**
image

**New Prompt**
image

**_Note: After a few tests, the text `Provide the most possible solution` seemed to be better than  `Provide the most likely solution` as it explored multiple options as opposed to focusing the answer on a single option._**

## ✅ Checks


pkg/ai/prompts.go
```diff
@@ -1,5 +1,9 @@
 package ai
 
 const (
-	default_prompt = "Simplify the following Kubernetes error message and provide a solution in %s: %s"
+	default_prompt = `Simplify the following Kubernetes error message delimited by triple dashes written in --- %s --- language; --- %s ---.
+	Provide the most possible solution in a step by step style in no more than 280 characters. Write the output in the following format:
+	Error:
```
matthisholleville, line 6: Is it really necessary to recall the error?

matthisholleville:
I wonder how the prompt behaves in case there are several errors on an object? Could you please provide an example with this case?

pkg/ai/prompts.go
camigira: Specifying this pattern for the output aims to have an explanation of the error. Happy to remove this if there's no value or we want to go with the idea of multiple prompts with different styles I proposed in Slack (for example non-technical, non-infra knowledge, etc.).

camigira:
> 
Example for a broken pod and deployment with multiple issues (hope this is what you had in mind with your question)

image

**broken-pod.yml**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: broken-pod
spec:
  containers:
    - name: broken-pod
      image: nginx:1.a.b.c
      livenessProbe:
        httpGet:
          path: /
          port: 90
        initialDelaySeconds: 3
        periodSeconds: 3
```

**broken-deployment.yml**
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: broken-deployment
spec:
  replicas: 3
  selector:
    matchLabels:
      app: myapp
  template:
    metadata:
      labels:
        app: myapp
    spec:
      containers:
        - name: nginx-container
          image: nginx:latest
          ports:
            - containerPort: 80
          livenessProbe:
            httpGet:
              path: /healthz #doesn't exist in nginx
              port: 80
            initialDelaySeconds: 30
            periodSeconds: 10
          readinessProbe:
            httpGet:
              path: /readiness #doesn't exist in nginx
              port: 80
            initialDelaySeconds: 20
            periodSeconds: 5
```

matthisholleville:
You still have only one error per object. 

Please try with this object: 

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: fake-ingress
spec:
  rules:
  - host: fake-ingress.example.com
    http:
      paths:
      - backend:
          service:
            name: fake-service
            port:
              number: 80
        path: /
        pathType: Prefix
  tls:
  - hosts:
    - fake-ingress.example.com
    secretName: fake-secret
```

camigira:
Thanks @matthisholleville for providing the example, here the results:

**Current Default Prompt**
```bash
0 default/fake-ingress(fake-ingress)
- Error: Ingress default/fake-ingress does not specify an Ingress class.
- Error: Ingress uses the service default/fake-service which does not exist.
- Error: Ingress uses the secret default/fake-secret as a TLS certificate which does not exist.
The Ingress resource is trying to use a service and a TLS certificate that do not exist, and it also does not specify which Ingress class to use.

To fix this error, create the necessary service and secret resources, and specify the Ingress class to use in the Ingress resource.
```

**Proposed Default Prompt**
```bash
0 default/fake-ingress(fake-ingress)
- Error: Ingress default/fake-ingress does not specify an Ingress class.
- Error: Ingress uses the service default/fake-service which does not exist.
- Error: Ingress uses the secret default/fake-secret as a TLS certificate which does not exist.
Error: Ingress default/fake-ingress needs an Ingress class, service default/fake-service doesn't exist, and secret default/fake-secret is a non-existent TLS cert.

Solution: 
1. Create a service named default/fake-service.
2. Create a secret named default/fake-secret with your TLS cert.
3. Add an ingress class to your Ingress resource, like nginx or traefik. 
4. Update your Ingress resource to use the service and secret created in step 1 and 2.
```

matthisholleville:
I really like the solution part @camigira but the error part in the prompt seems redondant.

What do you think @AlexsJones ?

AlexsJones (approved):

AlexsJones:
> I really like the solution part @camigira but the error part in the prompt seems redondant.
> 
> What do you think @AlexsJones ?

I think as long as it yields the outputs that are clear and understandable to the user its good. 
I don't know whether the error being specified in the block improves the prompt at all - we need to try with additional backends to be sure.

camigira:
👋 @AlexsJones @matthisholleville 

I saw this PR was approved but there are still unresolved conversations. Should I mark as completed and merge?

Thanks!

Issue #424, opened 2023-05-16
[Question]: Azure docs are out of date


AlexsJones:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### Steps to reproduce

```
k8sgpt auth --backend azureopenai --baseurl https://k8sgpt.openai.azure.com --engine k8sgpt-ai --model text-davinci-003
Error: unknown flag: --backend
```

### Expected behaviour

Should work for `new`

Issue #423, opened 2023-05-16
[Feature]: Acknowledgement on deletion of backend


AlexsJones:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've discussed this feature request in the [K8sGPT Slack](https://k8sgpt.slack.com) and got positive feedback

### Is this feature request related to a problem?

No

### Solution Description

It would be nice to say something like "Done" or "Backend %s removed"

<img width="543" alt="Screenshot 2023-05-16 at 07 06 22" src="https://github.com/k8sgpt-ai/k8sgpt/assets/1235925/8d85a71f-86c8-4455-99e7-1ea309886828">

### Benefits

Clarity

rakshitgondwal:
I can take this up.

AlexsJones:
Apologies but I have already started it as a broader piece of work

rakshitgondwal:
Cool

PR #427, opened 2023-05-16
feat: added the ability to set a user default AI provider

AlexsJones:
Closes # 

## ✅ Checks

- [x ] My pull request adheres to the code style of this project
- [ x] My code requires changes to the documentation
- [ x] I have updated the documentation as required
- [ x] All the tests have passed

## ℹ Additional Information

#423 #424

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -56,6 +56,7 @@ const (
 )
 
 type JsonOutput struct {
+	Provider string          `json:"provider"`
 	Errors   AnalysisErrors  `json:"errors"`
 	Status   AnalysisStatus  `json:"status"`
 	Problems int             `json:"problems"`
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -34,16 +34,17 @@ import (
 )
 
 type Analysis struct {
-	Context        context.Context
-	Filters        []string
-	Client         *kubernetes.Client
-	AIClient       ai.IAI
-	Results        []common.Result
-	Errors         []string
-	Namespace      string
-	Cache          cache.ICache
-	Explain        bool
-	MaxConcurrency int
+	Context            context.Context
+	Filters            []string
+	Client             *kubernetes.Client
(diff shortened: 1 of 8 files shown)


pkg/analysis/analysis.go
```diff
@@ -74,6 +75,12 @@ func NewAnalysis(backend string, language string, filters []string, namespace st
 		os.Exit(1)
 	}
 
+	// Backend string will have high priority than a default provider
+	// Backend as "openai" represents the default CLI argument passed through
+	if configAI.DefaultProvider != "" && backend == "openai" {
```
matthisholleville, line 80: Don't understand the use of this condition. Can you explain me please ?

pkg/analysis/output.go
```diff
@@ -56,6 +56,10 @@ func (a *Analysis) jsonOutput() ([]byte, error) {
 
 func (a *Analysis) textOutput() ([]byte, error) {
 	var output strings.Builder
+
+	// Print the AI provider used for this analysis
+	output.WriteString(fmt.Sprintf("AI Provider: %s\n", color.YellowString(a.AnalysisAIProvider)))
```
matthisholleville, line 61: Shouldn't we add this information in json output ?

matthisholleville (commented):
Great work. Left two comments

pkg/analysis/output.go
AlexsJones: Yeah I can do that!

pkg/analysis/analysis.go
AlexsJones: This means; if the user has set a DefaultProvider in their config use that.
The `Backend` will always default to `openai` even when they don't pass it through. 

There might be a case like this
> DefaultProvider: azureopenai

// Pass backend analyze --openai

In this case it will use openai

matthisholleville: Make sense.

AlexsJones:
Can you approve it @matthisholleville ?

Issue #209, opened 2023-04-04
feature: Support for remote cache


ArthurSens:
Checklist:
* [X] I've searched for similar issues and couldn't find anything matching
* [ ] I've discussed this feature in the #k8sgpt slack channel

## Is this feature request related to a problem?
* [ ] Yes
* [X] No

## Describe the solution you'd like

It would be nice if k8sgpt had the option to cache remotely.

## Benefits for the project and its users

With multiple individuals responsible for the same cluster/clusters, e.g. employees in the same company, a remote cache can decrease costs and latency if all those employees use the same cache.

matthisholleville:
I would like to propose a first implementation. Have you started something about this @AlexsJones ?

AlexsJones:
I haven't started anything, can you outline what you're thinking?

matthisholleville:
I think this issue needs to be split into two feature requests:
- Adding a lock file (similar to Terraform state file) during analysis.
- Adding remote storage functionality to store configuration files in S3 buckets, for example.

What do you think?

AlexsJones:
I think S3 can be set to do the locking for you?

matthisholleville:
Thank you for your answer. It's a good question to ask. From my understanding we need a lock or mechanism to ensure that the cache is not corrupted when parallelizing the analyses.

AlexsJones:
I am doing some work on designing this, will update here when I have something we can discuss further.

xavpaice:
Not sure if it's helpful, but you might want to see https://github.com/replicatedhq/troubleshoot, which has a `support-bundle` command that grabs a bunch of info from a running k8s cluster and pops it into a tarball, then https://github.com/replicatedhq/sbctl which has a utility capable of running a k8s api against that tarball.  I tried it with k8sgpt just now and there's absolutely some gaps, we need to collect more info and fix some things in sbctl that explode when k8sgpt queries it, but that's do-able.

The short story is, you can grab a totally disconnected tarball from a cluster, ship it somewhere, and then serve it up and run kubectl commands against it without having to connect to the live cluster.  Handy if you're doing disconnected support work.  I'd love to get k8sgpt to read those bundles.

AlexsJones:
[...]

PR #439, opened 2023-05-18
feat: caching

AlexsJones:
Closes # 

## 📑 Description
- Refactor README.md to reduce the amount of sprawl of information
- Introduce s3 based caching support #381 #209 

## ✅ Checks

### DIFF
--- a/cmd/cache/add.go
+++ b/cmd/cache/add.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cache
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/fatih/color"
+	"github.com/k8sgpt-ai/k8sgpt/pkg/cache"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// addCmd represents the add command
+var addCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a remote cache",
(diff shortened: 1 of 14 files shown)


cmd/cache/add.go
```diff
@@ -0,0 +1,71 @@
...
+			os.Exit(1)
+		}
+	},
+}
+
+func init() {
+	CacheCmd.AddCommand(addCmd)
+	addCmd.Flags().StringVarP(&region, "region", "r", "", "The region to use for the cache")
+	addCmd.Flags().StringVarP(&bucketname, "bucket", "b", "", "The name of the bucket to use for the cache")
+	addCmd.MarkFlagRequired("bucket")
```
matthisholleville, line 69: `region` is not required too ?

cmd/cache/remove.go
```diff
@@ -0,0 +1,60 @@
...
+			color.Red("Error: %v", err)
+			os.Exit(1)
+		}
+
+	},
+}
+
+func init() {
+	CacheCmd.AddCommand(removeCmd)
+	removeCmd.Flags().StringVarP(&bucketname, "bucket", "b", "", "The name of the bucket to use for the cache")
```
matthisholleville, line 59: Cela ne devrait-il pas être plus cohérent avec les autres commandes similaires telles que filters, auth, etc. ?

`k8sgpt cache remove my-bucket1,my-bucket2,...`

pkg/cache/file_based.go
```diff
@@ -19,6 +19,25 @@ func (f *FileBasedCache) IsCacheDisabled() bool {
 	return f.noCache
 }
 
+func (*FileBasedCache) List() ([]string, error) {
```
matthisholleville, line 22: What is the purpose of this function? Apart from knowing the name of the files, I don't see the value it will have for the user. What do you think about it?

matthisholleville (commented):
Very good work @AlexsJones !!!

Left some comments

pkg/cache/file_based.go
AlexsJones: This is eventually going to be built out to let you delete parts of the cache, for now it's just a utility to see what's in there

cmd/cache/add.go
AlexsJones: Yep done!

cmd/cache/remove.go
AlexsJones: Okay sure, will update

PR #450, opened 2023-05-22
feat: add update to auth cmd

### LINKED ISSUE #388: [Feature]: Enhance Auth command
### Checklist

No

### Problem Description

Auth could allow the user to update a previously configured provider. At the moment "auth" allows to list providers or add a new one.
Also, there's no background check on the validity of a configured key.

### Solution Description

- add an _update_ subcommand to the auth command
- Run a background check on the the key provided towards configured provider

### Benefits

- Allow to modify a configured provider
- Make sure a provided key is valid if not generated through _generate_ command

### Potential Drawbacks

_No response_

### Additional Information

_No response_

HatiCode:
Closes #388

## 📑 Description
This PR adds an update subcommand to the auth command, allowing to update a key in a previously configured provider e.g: password, model...
It also fixes a couple typos met along the way.

## ✅ Checks


matthisholleville:
Hello, thank you very much for your contribution. I just reviewed your PR. It seems like it implements an update method that requires setting all the fields. Imagine if the user wants to modify only the password in their configuration or the model without having to fill in all the other fields. I think it would be interesting to lean more towards this implementation.

What do you think ?

HatiCode:
@matthisholleville indeed this is what I'm trying to achieve, now because I reused the viper.Set() method at line 74, it reimplements all default empty values from the flags declared alongside the changes one. Do you know how I could keep the change passed with a flag and leave the rest untouched ?

matthisholleville:
Check if the flag is not empty? I don't think flags should have default values in the case of an update.

HatiCode:
@matthisholleville I made a couple changes and it's now working, changes are only applied for the flags used. I still feel like this could use some nice refactoring at a certain point tho

matthisholleville:
Hello @HatiCode can you please provide an example of the commande ?

PR #454, opened 2023-05-23
feat: support arbitrary uid for openshift environments

### LINKED ISSUE #440: [Bug]: k8sgpt fails to start with "mkdir /.config: permission denied"
### Checklist

### Affected Components

- [ ] K8sGPT (CLI)
- [X] #470

### K8sGPT Version

v0.3.0

### Kubernetes Version

v1.24.0

### Host OS and its Version

_No response_

### Steps to reproduce

Environment: OpenShift 4.11 with SCC enabled

Steps to reproduce:

1. install k8sgpt on K8s cluster with operator like described in https://github.com/k8sgpt-ai/k8sgpt-operator
2. provide secret and K8sGPT CR described in https://github.com/k8sgpt-ai/k8sgpt-operator#run-the-example
3. k8sgpt fails to start. Logs show as follows
```
kubectl logs k8sgpt-deployment-877c6ddd9-2cx5z
Error: mkdir /.config: permission denied
```

The pod runs by default with "restricted-v2" SCC (https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html). I guess this could be the reason.

securityContext of k8sgpt pod:

```
  securityContext:
    fsGroup: 1000980000
    seLinuxOptions:
      level: s0:c31,c25
    seccompProfile:
      type: RuntimeDefault
```

and securityContext of the container:

```
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsNonRoot: true
      runAsUser: 1000980000
```

### Expected behaviour

K8sGPT should run also on OpenShift with SCC enabled.

### Actual behaviour

K8sGPT fails to start on OpenShift 4.11 with SCC enabled.

### Additional Information

_No response_

jkleinlercher:
Closes #440

## 📑 Description
with this PR k8sgpt should also be able to run with an arbitrary uid. that is necessary especially in openshift environments because there by default the restricted-v2 SCC sets a uid in the securityContext. currently it depends on uid 65532 because k8sgpt wants to create a .config folder in the users home dir (https://github.com/k8sgpt-ai/k8sgpt/blob/6479cbaf919f25da08edaeb6440d7a3ff1ac9696/cmd/root.go#L88).

since only uid 65532 has a home dir defined in the distroless image an arbitrary uid would want to create the .config dir in the root folder, which is not allowed. so with this PR the XDG_ variables are pointing to a different folder which gets mounted as an emptyDir volume.

## ✅ Checks


AlexsJones:
The code looks good but I have no ability to test this, @jkleinlercher does this work on OS environments?

AlexsJones (approved):

PR #457, opened 2023-05-23
feat: get official field doc

### LINKED ISSUE #456: [Feature]: get official field doc
### Checklist

No

### Problem Description

It could be useful to also get the official doc for the field in error

### Solution Description

A simple http request to the k8S api reference page and parsing the result.

### Benefits

A better understand of the issue

### Potential Drawbacks

_No response_

### Additional Information

_No response_

golgoth31:
Closes #456

## 📑 Description
Make an http call to the oficial k8S api reference page to get field documentation

## ✅ Checks

## ℹ Additional Information
This is currently a first preview to be discussed.


golgoth31:
a simple exemple to show what it could looks like
before (from version v0.3.4):
image

after (from main):
image

matthisholleville:
Thank you for your contribution. I think it is an important and interesting information to display to the user. What do you think @AlexsJones ?

Can you test to deploy the following object and check that in case an object returns several errors, the information remains digestible please?

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: fake-ingress
spec:
  rules:
  - host: fake-ingress.example.com
    http:
      paths:
      - backend:
          service:
            name: fake-service
            port:
              number: 80
        path: /
        pathType: Prefix
  tls:
  - hosts:
    - fake-ingress.example.com
    secretName: fake-secret
```

You should have this result with actual version :

```bash
0 k8sgpt/fake-ingress(fake-ingress)
- Error: Ingress k8sgpt/fake-ingress does not specify an Ingress class.
- Error: Ingress uses the service k8sgpt/fake-service which does not exist.
- Error: Ingress uses the secret k8sgpt/fake-secret as a TLS certificate which does not exist.
```

golgoth31:
Hi,
with the actual version, I have:
0 default/fake-ingress(fake-ingress)
- Error: Ingress default/fake-ingress does not specify an Ingress class.
- Error: Ingress uses the service default/fake-service which does not exist.
- Error: Ingress uses the secret default/fake-secret as a TLS certificate which does not exist.
Error: Ingress default/fake-ingress doesn't specify an Ingress class. Ingress uses the service default/fake-service, which doesn't exist. Ingress uses the secret default/fake-secret as a TLS certificate, which doesn't exist.

Solution: 
1. Create the fake-service & fake-secret objects.
2. Add annotation "kubernetes.io/ingress.class: [class-name]" to fake-ingress definition.
3. Apply the changes to the cluster.

with the patch, I have:
0 default/fake-ingress(fake-ingress)
- Error: Ingress default/fake-ingress does not specify an Ingress class.
  Official Doc: ingressClassName is the name of an IngressClass cluster resource. Ingress controller implementations use this field to know whether they should be serving this Ingress resource, by a transitive connection (controller -> IngressClass -> Ingress resource). Although the `kubernetes.io/ingress.class` annotation (simple constant name) was never formally defined, it was widely supported by Ingress controllers to create a direct binding between Ingress controller and Ingress resources. Newly created Ingress resources should prefer using the field. However, even though the annotation is officially deprecated, for backwards compatibility reasons, ingress controllers should still honor that annotation if present.
- Error: Ingress uses the service default/fake-service which does not exist.
- Error: Ingress uses the secret default/fake-secret as a TLS certificate which does not exist.
Error: Ingress default/fake-ingress does not specify an Ingress class.
Solution: 
1. Define an Ingress class using the ingressClassName field in your YAML file.
2. If needed for backwards compatibility, add the kubernetes.io/ingress.class annotation.
3. Make sure the Ingress is referencing a valid service and secret.

matthisholleville:
Thank you. Shoudn't we have `Official Doc` part for each errors ?

golgoth31:
I only extract the higher level field. Il have to find à way to parse also the subfields from the doc

golgoth31:
I have an other way to find the information, I'll made some tests ... :)

golgoth31:
Hi,
I have modified the patch to get the schema from the current server and extract data from it. The result is like this:
```
1 default/fake-ingress(fake-ingress)
- Error: Ingress default/fake-ingress does not specify an Ingress class.
  Kubernetes Doc: ingressClassName is the name of an IngressClass cluster resource. Ingress controller implementations use this field to know whether they should be serving this Ingress resource, by a transitive connection (controller -> IngressClass -> Ingress resource). Although the `kubernetes.io/ingress.class` annotation (simple constant name) was never formally defined, it was widely supported by Ingress controllers to create a direct binding between Ingress controller and Ingress resources. Newly created Ingress resources should prefer using the field. However, even though the annotation is officially deprecated, for backwards compatibility reasons, ingress controllers should still honor that annotation if present.
- Error: Ingress uses the service default/fake-service which does not exist.
  Kubernetes Doc: service references a service as a backend. This is a mutually exclusive setting with "Resource".
- Error: Ingress uses the secret default/fake-secret as a TLS certificate which does not exist.
  Kubernetes Doc: secretName is the name of the secret used to terminate TLS traffic on port 443. Field is left optional to allow TLS routing based on SNI hostname alone. If the SNI host in a listener conflicts with the "Host" header field used by an IngressRule, the SNI host is used for termination and value of the "Host" header is used for routing.
```

I've added a new field in the common.Failure struct to store the kubernetes doc string.
Once the kind, group and discovery client are set in the kubernetes.K8sApiReference struct, any field can be retreived by getting the path (eg: "spec.rules.http.paths.backend.service")

matthisholleville:
Thank you very much @golgoth31 ! We really appreciate the contribution and we could use it as option with `--with-doc` flag on analysis. Can you please add flag to use this option ?

golgoth31:
> Thank you very much @golgoth31 ! We really appreciate the contribution and we could use it as option with `--with-doc` flag on analysis. Can you please add flag to use this option ?

Thank you,
I've added the with-doc flag, to ease the use of the flag, I move the getting of the openapi schema in the runanalisys function. 
I enforced the with-doc to false in server mode. I can make a PR to the protobuf if needed.

pkg/analysis/output.go
```diff
@@ -78,6 +78,9 @@ func (a *Analysis) textOutput() ([]byte, error) {
 			color.YellowString(result.Name), color.CyanString(result.ParentObject)))
 		for _, err := range result.Error {
 			output.WriteString(fmt.Sprintf("- %s %s\n", color.RedString("Error:"), color.RedString(err.Text)))
+			if err.KubernetesDoc != "" {
```
matthisholleville, line 81: Shouldn't we add this on `json` output too ?

pkg/kubernetes/apireference.go
```diff
@@ -0,0 +1,77 @@
...
+	openapi_v2 "github.com/google/gnostic/openapiv2"
+)
+
+func (k *K8sApiReference) GetApiDocV2(field string) string {
+	startPoint := ""
+	// the path must be formated like "path1.path2.path3"
+	paths := strings.Split(field, ".")
+	group := strings.Split(k.ApiVersion.Group, ".")
+
+	// openapiSchema := openapi_v2.Document{}
```
matthisholleville, line 15: Shouldn't we remove this ?

pkg/kubernetes/apireference.go
```diff
@@ -0,0 +1,77 @@
...
+	// openapiSchema, err := k.Discovery.OpenAPISchema()
+	// if err != nil {
+	// 	return ""
+	// }
+
+	definitions := k.OpenapiSchema.GetDefinitions().GetAdditionalProperties()
+
+	// extract the startpoint by searching the highest leaf corresponding to the requested group qnd kind
+	for _, prop := range definitions {
+		if strings.HasSuffix(prop.GetName(), group[0]+"."+k.ApiVersion.Version+"."+k.Kind) {
```
matthisholleville, line 26: Can we use `fmt.Sprintf` to be more consistent ?

matthisholleville (changes requested):
Left few comments. Good job ! Can you please add usage example in README please ?

pkg/analysis/output.go
golgoth31: The kubernetesDoc is a filed of the failure struct, and so, part of the common.Result struct. At the end it's already in the json output.
```json
{
  "provider": "openai",
  "errors": null,
  "status": "ProblemDetected",
  "problems": 4,
  "results": [
    {
      "kind": "Ingress",
      "name": "default/fake-ingress",
      "error": [
        {
          "Text": "Ingress default/fake-ingress does not specify an Ingress class.",
          "KubernetesDoc": "ingressClassName is the name of an IngressClass cluster resource. Ingress controller implementations use this field to know whether they should be serving this Ingress resource, by a transitive connection (controller -\u003e IngressClass -\u003e Ingress resource). Although the `kubernetes.io/ingress.class` annotation (simple constant name) was never formally defined, it was widely supported by Ingress controllers to create a direct binding between Ingress controller and Ingress resources. Newly created Ingress resources should prefer using the field. However, even though the annotation is officially deprecated, for backwards compatibility reasons, ingress controllers should still honor that annotation if present.",
          ...
```

pkg/kubernetes/apireference.go
golgoth31: yes, of course

pkg/kubernetes/apireference.go
golgoth31: fixed

matthisholleville (approved):
LGTM.

Another review @AlexsJones ?

golgoth31:
The readme has been updated

AlexsJones (approved):

AlexsJones:
This is a great first-time contribution thanks @golgoth31

PR #461, opened 2023-05-25
fix: optimize analyze service

cleverhu:
Closes # 

## 📑 Description

When I used the analysis command, I found that the yellow error 'Service kube system/cluster. local nfs client nfs sub dir external advisor does not exist' was reported such as:
uTools_1685028836704
However, during my investigation, I found that the purpose of this endpoint is not to discover the pod IP, but to act as a leaderAction. Therefore, I believe this should not be thrown as a yellow warning.
uTools_1685029003724

## ✅ Checks


matthisholleville:
Hello @cleverhu, thank you very much for your contribution. Do we really need to have this information?

Shouldn't we just proceed ( `continue` ) if it matches your condition?

cleverhu:
> Hello @cleverhu, thank you very much for your contribution. Do we really need to have this information?
> 
> 
> 
> Shouldn't we just proceed ( `continue` ) if it matches your condition?

Yes, I think it is correct to omit any debugging information from the output.

AlexsJones:
Yay another first time contributor - welcome!

Is this read for review again?

PR #483, opened 2023-06-06
feat: analyze specific resource(s)

### LINKED ISSUE #285: feature: add diagnose command
Checklist:

* [ ] Yes
* [x] No

Currently k8sgpt analyzes a list of resources in a Kubernetes cluster. It is not possible to separate an object.

We should to add the possibility to diagnose specific ressource like `k8sgpt diagnose pod my-specific-pod`

This feature would allow the user to be more precise in what he wants to analyze.

panpan0000:
Closes #285 

## 📑 Description

add capability to analyze specific resource(s)

## ✅ Checks

### Design Overview and Demo:
The usage is similar as `kubectl get xxx` ( one resource or multiple resource, but should be in the same namespaces(for namespaced-scope resources))
But note: it doesn't support k8s short-name so far ( like `service` -> `svc`).

### Example 1:
```
k8sgpt  analyze -n default  pod/deployment003-74db995cbd-m2t8x
```
image

### Example 2:

just like kubectl can do `kubectl get deployment/deployment003  service/mysql-service`, 

multiple resources are analyzed

```
k8sgpt  analyze deployment/deployment003  service/mysql-service
```
image

### Example 3:

co-work with filter: `filter` takes priority and find the overlap.

```
k8sgpt  analyze  deployment/deployment003  service/mysql-service --filter=Deployment
```

image

### Example 4:

specific resource not existing

image

### More

add example in `-h`
image

### Additional Changes:

- when `no namespace` is given, "default" will be filled automatically.
- add example in cmd helper

### DIFF
--- a/cmd/cache/list.go
+++ b/cmd/cache/list.go
@@ -42,7 +42,7 @@ var listCmd = &cobra.Command{
 			color.Red("Error: %v", err)
 			os.Exit(1)
 		}
-		
+
 		for _, name := range names {
 			println(name)
 		}
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -47,6 +47,7 @@ type Analysis struct {
 	MaxConcurrency     int
 	AnalysisAIProvider string // The name of the AI Provider used for this analysis
 	WithDoc            bool
+	SpecificResources  []string // format: slices of <kind>/<name> or just <name>
 }
 
 type AnalysisStatus string
@@ -65,7 +66,7 @@ type JsonOutput struct {
 	Results  []common.Result `json:"results"`
 }
 
-func NewAnalysis(backend string, language string, filters []string, namespace string, noCache bool, explain bool, maxConcurrency int, withDoc bool) (*Analysis, error) {
+func NewAnalysis(backend string, language string, filters []string, namespace string, noCache bool, explain bool, maxConcurrency int, withDoc bool, specificResources []string) (*Analysis, error) {
 	var configAI ai.AIConfiguration
 	err := viper.UnmarshalKey("ai", &configAI)
 	if err != nil {
@@ -131,6 +132,7 @@ func NewAnalysis(backend string, language string, filters []string, namespace st
 		MaxConcurrency:     maxConcurrency,
 		AnalysisAIProvider: backend,
 		WithDoc:            withDoc,
+		SpecificResources:  specificResources,
 	}, nil
 }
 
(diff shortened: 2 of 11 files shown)


panpan0000:
@matthisholleville  @AlexsJones  sorry for late for Mat's invitation long time ago, here comes the code :-)

panpan0000:
not a push ,but if any question, let me know please :-)

pkg/analysis/analysis.go
```diff
@@ -190,6 +253,12 @@ func (a *Analysis) RunAnalysis() {
 		var wg sync.WaitGroup
 		var mutex sync.Mutex
 		for _, filter := range a.Filters {
+			if len(analyzerConfig.Resources) > 0 {
+				if _, ok := analyzerConfig.Resources[filter]; !ok {
+					// NOTE: if specific resources given(example: pod/nginx), the other analyzer(like deploy/svc) can be skipped
```
bwplotka, line 258:
```suggestion
					// A resource of a kind that this analyzer does not support was specified, skipping.
```

pkg/analysis/analysis.go
```diff
@@ -219,6 +288,12 @@ func (a *Analysis) RunAnalysis() {
 	semaphore = make(chan struct{}, a.MaxConcurrency)
 	// use active_filters
 	for _, filter := range activeFilters {
+		if len(analyzerConfig.Resources) > 0 {
+			if _, ok := analyzerConfig.Resources[filter]; !ok {
+				// NOTE: if specific resources given(example: pod/nginx), the other analyzer(like deploy/svc) can be skipped
```
bwplotka, line 293: ditto

pkg/analyzer/deployment.go
```diff
@@ -45,14 +47,25 @@ func (d DeploymentAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error)
 	AnalyzerErrorsMetric.DeletePartialMatch(map[string]string{
 		"analyzer_name": kind,
 	})
-
-	deployments, err := a.Client.GetClient().AppsV1().Deployments(a.Namespace).List(context.Background(), v1.ListOptions{})
-	if err != nil {
-		return nil, err
+	var list *appsv1.DeploymentList = &appsv1.DeploymentList{}
+	if len(a.Resources["Deployment"]) > 0 {
```
bwplotka, line 51: Can we return error on all other analyzers that does not yet support "resources"? 🤔 otherwise it might be bit misleading.

bwplotka (commented):
Nice work! I like this feature.

Suggested some code/style changes. 

The CLI might work in this form, but there is a duplicate with filters indeed. It's also a new pattern as previously all other parameters was provided by flags. Given this is what kubectl is supporting, the proposed form could work.

One thing is that all other analyzers which does not support yet resources - they probably should return error if they are responsible for "kind" but does not support "resource" specification 🤔 

Otherwise it will be frustrating to say `cronjob/<abc>` and cronjob analyzer will scan ALL cron jobs. Fast fail here would be expected perhaps?

cmd/cache/list.go
```diff
@@ -42,7 +42,7 @@ var listCmd = &cobra.Command{
 			color.Red("Error: %v", err)
 			os.Exit(1)
 		}
-		
+
 		for _, name := range names {
```
bwplotka, line 46: Rebase is needed I beliece 🤔

bwplotka:
Another issue (could be done in another PR, but ticket would be nice for it): If I specify `<KIND1>/<name>` and `kind1` is not a valid filter OR it's unused filter, we could fail fast.

NOTE: I am using your amazing feature on my [fork](https://github.com/bwplotka/k8sgpt/tree/gmp-podmonitoring) for some experimentations, thanks for this! Looking forward to having it in upstream 💪🏽

pkg/analysis/analysis.go
```diff
@@ -157,13 +207,26 @@ func (a *Analysis) RunAnalysis() {
 		AIClient:      a.AIClient,
 		OpenapiSchema: openapiSchema,
 	}
+	analyzerConfig.Resources = make(map[string][]string)
+	for k, v := range resourceMap {
+		analyzerConfig.Resources[k] = v
+	}
+	if len(analyzerConfig.Resources) > 0 && len(analyzerConfig.Namespace) == 0 {
```
bwplotka, line 214: Why not always set it to "default" on empty? Also comment is probably not needed here.

bwplotka: nvm, I am wrong here.

AlexsJones:
We took a while to review this @panpan0000 I am sorry for that. Do you still want to make the changes and have us merge it?

bwplotka:
Now when play with this more, it feels having `--filter` understand resource name (not only kind) would be also a good API (naturally working)

panpan0000:
> Cherry picked and rebased (without style fixes)

go ahead @bwplotka ,  very appreciated this idea is helpful and you can finish this work. 
I don't know  people will revisit this half an year later.... ^_^
Thank you very much @bwplotka  , sorry that too busy with other stuff...

I will close this PR once your done. Feel free to do anything based on my commit or code, if they helps it's a great joy.

PR #553, opened 2023-07-13
feat: admission webhooks

AlexsJones:
Closes # 

## ✅ Checks


pkg/analyzer/mutating_webhook.go
```diff
@@ -0,0 +1,133 @@
...
+	var preAnalysis = map[string]common.PreAnalysis{}
+
+	for _, webhookConfig := range mutatingWebhooks.Items {
+		for _, webhook := range webhookConfig.Webhooks {
+			var failures []common.Failure
+
+			svc := webhook.ClientConfig.Service
+			// Get the service
+			service, err := a.Client.GetClient().CoreV1().Services(svc.Namespace).Get(context.Background(), svc.Name, v1.GetOptions{})
+			if err != nil {
```
arbreezy, line 59: @AlexsJones I think checking if err is not found and produce an error like "this service doesn't exist for this webhook"  is also something we can have in the analysis

AlexsJones: Good point

pkg/analyzer/mutating_webhook.go
```diff
@@ -57,6 +57,17 @@ func (MutatingWebhookAnalyzer) Analyze(a common.Analyzer) ([]common.Result, erro
 			// Get the service
 			service, err := a.Client.GetClient().CoreV1().Services(svc.Namespace).Get(context.Background(), svc.Name, v1.GetOptions{})
 			if err != nil {
+				// If the service is not found, we can't check the pods
+				failures = append(failures, common.Failure{
+					Text:          fmt.Sprintf("Service %s not found as mapped to by Mutating Webhook %s", svc.Name, webhook.Name),
+					KubernetesDoc: apiDoc.GetApiDocV2("spec.webhook.clientConfig.service"),
+					Sensitive: []common.Sensitive{
```
arbreezy, line 64: I believe masking the service name also makes sense here

pkg/analyzer/mutating_webhook.go
```diff
@@ -0,0 +1,148 @@
...
+	var preAnalysis = map[string]common.PreAnalysis{}
+
+	for _, webhookConfig := range mutatingWebhooks.Items {
+		for _, webhook := range webhookConfig.Webhooks {
+			var failures []common.Failure
+
+			svc := webhook.ClientConfig.Service
+			// Get the service
+			service, err := a.Client.GetClient().CoreV1().Services(svc.Namespace).Get(context.Background(), svc.Name, v1.GetOptions{})
+			if err != nil {
```
arbreezy, line 59: It may be minor but is better to explicitly check the `errors.IsNotFound(err)` so we can separate the failures block code from other errors from the k8s client

AlexsJones: Sure lets do it

arbreezy (approved):

PR #621, opened 2023-08-31
chore: fixes a bug where filters do not deactivate

AlexsJones:
Closes # 

## ✅ Checks

- [ x] My pull request adheres to the code style of this project
- [ x] My code requires changes to the documentation
- [ x] I have updated the documentation as required
- [x ] All the tests have passed

### DIFF
--- a/pkg/integration/integration.go
+++ b/pkg/integration/integration.go
@@ -32,8 +32,6 @@ type IIntegration interface {
 	UnDeploy(namespace string) error
 	//
 	AddAnalyzer(*map[string]common.IAnalyzer)
-	// RemoveAnalyzer removes an analyzer from the cluster
-	RemoveAnalyzer() error
 
 	GetAnalyzerName() []string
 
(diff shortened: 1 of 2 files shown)


pkg/integration/integration.go
```diff
@@ -108,23 +104,13 @@ func (*Integration) Deactivate(name string, namespace string) error {
...
-	foundFilter := false
-	for i, v := range activeFilters {
-
-		for _, intanal := range integrations[name].GetAnalyzerName() {
-			if v == intanal {
-				foundFilter = true
-				activeFilters = append(activeFilters[:i], activeFilters[i+1:]...)
-				break
+	// Update filters and remove the specific filters for the integration
+	for _, filter := range integrations[name].GetAnalyzerName() {
```
matthisholleville, line 108: What happens if I attempt to update or remove an unactivated filter?

AlexsJones: This is true.. it's only parsing the active filters, will need to check the deactivated ones also, good spot

AlexsJones: I think this would be okay upon second thought. 
The unused filter is only listed from the integration filters, which would go once the integration is deactivated

AlexsJones:
In reverse order, ( activating the integration, disabling a filter, disabling integration, everything back to normal ) 
Screenshot 2023-09-03 at 16 12 19
Screenshot 2023-09-03 at 16 10 18
Screenshot 2023-09-03 at 16 11 06

PR #690, opened 2023-10-01
feat: add Azure remote cache

arbreezy:
Closes # 

## 📑 Description

PR adds a new remote cache, integrating with Azure Blob storage offering.
It also includes the new protobuf schema.

Finally, it reworks the cache code path to make it easier to introduce more remote cache types in the future 

## ✅ Checks


cmd/cache/add.go
```diff
@@ -46,9 +51,15 @@ var addCmd = &cobra.Command{
...
-	addCmd.MarkFlagRequired("region")
-
+	addCmd.Flags().StringVarP(&region, "region", "r", "", "The region to use for the AWS S3 cache")
+	addCmd.Flags().StringVarP(&bucketname, "bucket", "b", "", "The name of the AWS S3 bucket to use for the cache")
+	addCmd.MarkFlagsRequiredTogether("region", "bucket")
+	addCmd.Flags().StringVarP(&storageAccount, "storageacc", "s", "", "The Azure storage account name of the container")
+	addCmd.Flags().StringVarP(&containerName, "container", "c", "", "The Azure container name to use for the cache")
+	addCmd.MarkFlagsRequiredTogether("storageacc", "container")
+	// Tedious check to ensure we don't include arguments from different providers
+	addCmd.MarkFlagsMutuallyExclusive("region", "storageacc")
```
AlexsJones, line 61: I didnt know about this method!

AlexsJones (approved):
This has to be cautiously merged as it will break the caching functionality in the operator I believe @arbreezy

AlexsJones:
Should this get merged in after you do the operator work or now @arbreezy ?

pkg/cache/cache.go
```diff
@@ -7,6 +7,14 @@ import (
 	"github.com/spf13/viper"
 )
 
+type CacheType string
+
+const (
+	Azure     CacheType = "azure"
+	S3        CacheType = "s3"
+	FileBased CacheType = "file"
+)
```
matthisholleville, line 16: Shouldn't we handle cache providers in the same way we handle analyzers/filters?

```go
var remoteCacheProviderMap = map[string]ICache{
	"Azure":  &S3Cache{},
	"Azure": &AzureCache{},
        "File": &FileBasedCache{},
}

func ListRemoteCacheProviders() []string {
	remoteCacheProviderKeys := make([]string, 0, len(remoteCacheProviderMap))
	for k := range remoteCacheProviderMap {
		remoteCacheProviderKeys = append(remoteCacheProviderKeys, k)
	}
	return remoteCacheProviderKeys
}
```

matthisholleville (commented):
Very good MR ! Nice job Aris

pkg/cache/cache.go
arbreezy: a) I think there is not a big value out of having this as an interface, b) the list functionality in caches is listing files from your already configured cache backend and not a list of supported caches from.

AlexsJones (approved):

PR #851, opened 2024-01-08
feat: deep dive alerts

### LINKED ISSUE #795: [Feature]: Deep dive alerts
### Checklist

No

### Problem Description

When monitoring my Kubernetes clusters, I often receive alerts due to a pod being in a crash loop or other issues for example. Investigating these alerts requires repetitive tasks such as log analysis, pod description, etc. K8sgpt allows me to anticipate some alerts by continuously analyzing my objects, but could we do even better?

### Solution Description

Who hasn't dreamed of having a tool that can perform an initial investigation into one or more alerts during oncall rotation ? The solution I propose is to leverage OpenAI's assistant system (or a similar pattern) to transition from a proactive mode to a reactive mode specific to an alert. The architecture would be as follows:

1. AlertManager sends one or more alerts to the K8SGPT operator via http.
2. K8SGPT uses the [OpenAI assistant](https://platform.openai.com/docs/assistants/overview) system to determine which functions (or executors) it needs to execute to doing error analysis.
3. With the results of the functions (or executors), we conduct an analysis and provide the user with an initial investigation of the error.

### Benefits

The benefits would be multiple:

1. Improved on-call process: Alerts received at night (when we are not operational at 100%) are automatically investigated.
2. Reduction of repetitive tasks and the possibility of errors (accidentally running a command in the middle of the night).
3. Automatic initiation and drafting of a post-mortem based on the analysis.

### Potential Drawbacks

The drawbacks could be the compatibility of the solution with all AI systems supported by K8SGPT. Given that the concept of [OpenAI's assistant](https://platform.openai.com/docs/assistants/overview) has just been introduced, it needs to be verified whether this concept will become a standard in the future. Otherwise, the system may need to be "complexified" to use a pattern [router](https://python.langchain.com/docs/modules/chains/foundational/router). Another solution could be to offer this functionality exclusively to OpenAI users for the time being.

### Additional Information

I have already tested this solution, and it has provided significant value for simple alerts. If the idea seems promising, the next step would be to test it on more complex alerts.
demo-alert

matthisholleville:
Closes #795 

## :bookmark_tabs: Description

**Disclaimer** : This PR is a first implementation and should be consolidated

It's deeply reworks our interaction with the AI. Until now, we had an AI agent (which we could call "analyzer"). Its goal is to provide an explanation for an error sample. This agent is still used for the "classic" explain mode.

For "deep dive alert" analysis, it was necessary to transform this model to switch to a **multi-agent autonomous AI system**. 

This system consists of 3 AI agents:
1. `router`: its goal is to identify the appropriate analyzers to activate based on the provided alert (`k8sgpt analyze --alert ...`).
2. `evaluator`: once the analysis has been performed, this agent's objective is to filter errors related to the provided alert. This agent helps refine errors that may be related to the alert.
3. `analyzer`: it's the historical analyzer responsible for finding a solution to a problem.

A new `pkg` called `agent` has been created to host the code for these agents.

https://github.com/k8sgpt-ai/k8sgpt/assets/99146727/9b849079-9cf4-420f-b5f6-835e123a9c60

## :white_check_mark: Checks

## ℹ Additional Information

Other improvements:

- A loader has been added during analysis to enhance the user experience in CLI mode ( we should not display the loader if `json` format provided )
- A description has been added to each of the analyzers to provide context to the "router" agent about the purpose of the analyzer.

This PR is a first implementation and should be consolidated with the following elements:

- [ ] Modification of [k8sgpt-schemas](https://github.com/k8sgpt-ai/schemas) to add the concept of alerts.
- [ ] The prompts need to be tested with different alerts.
- [ ] The documentation needs to be updated.
- [ ] Test with other AI Provider ( tested with OpenAI )
- [ ] Do not display the loader if the `json` format is provided.

**With this mode, we make 2 additional AI API calls per command (router and evaluator agent).**

-------

Analysis flow with error 

image


cmd/analyze/analyze.go
```diff
@@ -44,24 +47,55 @@ var AnalyzeCmd = &cobra.Command{
 	provide you with a list of issues that need to be resolved`,
 	Run: func(cmd *cobra.Command, args []string) {
 
+		if alert != "" {
+			explain = true
```
AlexsJones, line 51: What if I dont want AI or alerts?

cmd/analyze/analyze.go
```diff
@@ -88,6 +122,8 @@ func init() {
 	AnalyzeCmd.Flags().BoolVarP(&nocache, "no-cache", "c", false, "Do not use cached data")
 	// anonymize flag
 	AnalyzeCmd.Flags().BoolVarP(&anonymize, "anonymize", "a", false, "Anonymize data before sending it to the AI backend. This flag masks sensitive data, such as Kubernetes object names and labels, by replacing it with a key. However, please note that this flag does not currently apply to events.")
+	// alert flag
+	AnalyzeCmd.Flags().StringVarP(&alert, "alert", "", "", "Alert to be analyzed")
```
AlexsJones, line 126: Let's offer a shorthand flag also, it doesn't need to be "-a"

pkg/agent/analyzer.go
```diff
@@ -0,0 +1,61 @@
...
+	default_prompt = `Simplify the following Kubernetes error message delimited by triple dashes written in --- %s --- language; --- %s ---.
+	Provide the most possible solution in a step by step style in no more than 280 characters. Write the output in the following format:
+	Error: {Explain error here}
+	Solution: {Step by step solution here}
+	`
+	trivy_vuln_prompt = "Explain the following trivy scan result and the detail risk or root cause of the CVE ID, then provide a solution. Response in %s: %s"
+	trivy_conf_prompt = "Explain the following trivy scan result and the detail risk or root cause of the security check, then provide a solution."
+)
+
+var PromptMap = map[string]string{
```
AlexsJones, line 21: Can we comment what this does?

AlexsJones:
What makes this OpenAI specific? I expected to see us use OpenAI functions, but it looks like we're just doing some smart prompt manipulation instead.

cmd/analyze/analyze.go
matthisholleville: This mode is necessarily associated with AI since the router and the evaluator agent are calls to the AI.

matthisholleville:
We had the choice between using OpenAI's [assistant](https://platform.openai.com/docs/assistants/how-it-works) mode and being specific to OpenAI, or using a multi-agent system that we maintain ourselves. This is the choice I made so that this mode is available for all AI backends.

AlexsJones:
> We had the choice between using OpenAI's [assistant](https://platform.openai.com/docs/assistants/how-it-works) mode and being specific to OpenAI, or using a multi-agent system that we maintain ourselves. This is the choice I made so that this mode is available for all AI backends.

 Okay that's even better. I am looking for an opportunity to re-use some of your work in the interactive loop I am making

cmd/analyze/analyze.go
matthisholleville: What do you think about "l" ?

matthisholleville:
If you need help or suggestions for integrating multi-agent functionality, reach out to me. 

I can finalize the few remaining jobs quickly to have a 'ready for review' pull request.

cmd/analyze/analyze.go
AlexsJones: that's fine, might be easier to make it a less ambiguous char like `g` or `y`

AlexsJones:
Sorry I lost this thread @matthisholleville - I think my primary reservation was was originally around it being bound to OpenAI functionality; I think you subsequently disproved that. Shall we get this through another review? The schema has just been approved.

matthisholleville:
@AlexsJones I'm diving back into it this week and I'll see about getting it ready for review as quickly as possible.

PR #855, opened 2024-01-10
feat: initial Prometheus analyzers

pintohutch:
Closes # 

## 📑 Description
Added a prometheus integration with two analyzers:
1. `PrometheusConfigValidate`
2. `PrometheusConfigRelabelReport`

The integration does not deploy any Prometheus stack in the cluster.
Instead, it searches the provided `--namespace` for a Prometheus
configuration, stored in a ConfigMap or Secret. If it finds one, it
unmarshals it into memory and runs the analyzers on it.

`PrometheusConfigValidate` checks if the actual Prometheus configuration is valid or has
any errors.

`PrometheusConfigRelabelReport` tries to distill the scrape config
relabeling rules to give a concise label set per job that targets need
to have to be scraped. This analyzer is unconventional, in that it does
not necessarily mean there are issues with the config. It merely tries
to give a human-readable explanation of the relabel rules it discovers,
leaning on the LLM and prompt.

Tested on both kube-prometheus and Google Managed Prometheus
stacks.

## ✅ Checks


pintohutch:
Also for the `PrometheusConfigRelabelReport` analyzer, the entire relabel config is inserted into the prompt. This may not be ideal and prints a verbose "error" message.

Maybe there's a better way to do this as well.

pkg/ai/prompts.go
```diff
@@ -8,10 +8,51 @@ const (
 	`
 	trivy_vuln_prompt = "Explain the following trivy scan result and the detail risk or root cause of the CVE ID, then provide a solution. Response in %s: %s"
 	trivy_conf_prompt = "Explain the following trivy scan result and the detail risk or root cause of the security check, then provide a solution."
+
+	prom_conf_prompt = `Simplify the following Prometheus error message delimited by triple dashes written in --- %s --- language; --- %s ---.
+	This error came from the Prometheus configuration file, documented at https://prometheus.io/docs/prometheus/latest/configuration/configuration/.
```
matthisholleville, line 13: I am thinking about the value of having the link in the prompt. The LLM will surely indicate the link in the response, and we might face confusion regarding the Prometheus version (e.g., my current Prometheus version may not be the latest, but the link indicates the latest). What do you think?

pkg/integration/prometheus/config_analyzer.go
```diff
@@ -0,0 +1,287 @@
...
+	return proms, nil
+}
+
+func findPrometheusConfigPath(ctx context.Context, client kubernetes.Interface, pod *corev1.Pod) (string, error) {
+	var path string
+	var err error
+	for _, container := range pod.Spec.Containers {
+		for _, arg := range container.Args {
+			if strings.HasPrefix(arg, prometheusConfigFlag) {
+				path = strings.TrimLeft(arg, prometheusConfigFlag)
```
matthisholleville, line 192: Shouldn't we end the for loop if we find the path ?

pkg/integration/prometheus/config_analyzer.go
```diff
@@ -0,0 +1,287 @@
...
+			}
+			if strings.HasPrefix(arg, configReloaderConfigFlag) {
+				path = strings.TrimLeft(arg, configReloaderConfigFlag)
```
matthisholleville, line 195: Same as the last comment here ?

pkg/integration/prometheus/config_analyzer.go
```diff
@@ -0,0 +1,287 @@
...
+	// Unmarshal the data into a Prometheus config.
+	if err := yaml.Unmarshal(b, &config); err == nil {
+		return &config, nil
+		// If there were errors, try gunziping the data.
+	} else if content := http.DetectContentType(b); content == "application/x-gzip" {
+		r, err := gzip.NewReader(bytes.NewBuffer(b))
+		if err != nil {
+			return &config, err
+		}
+		ub, err := io.ReadAll(r)
```
matthisholleville, line 275: Should we update `ub` to a more comprehensive variable ?

pkg/integration/prometheus/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+
+type Prometheus struct {
+}
+
+func NewPrometheus() *Prometheus {
+	return &Prometheus{}
+}
+
+func (p *Prometheus) Deploy(_ string) error {
+	// no-op
```
matthisholleville, line 21: We don't want to fallback to a Helm chart installation (as in the case of Trivy integration) if we don't detect Prometheus installed on the cluster, right?

pkg/integration/prometheus/relabel_analyzer.go
```diff
@@ -0,0 +1,85 @@
...
+type RelabelAnalyzer struct {
+}
+
+func (r *RelabelAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
+	ctx := a.Context
+	client := a.Client.GetClient()
+	namespace := a.Namespace
+	kind := ConfigRelabel
+
+	podConfigs, err := findPrometheusPodConfigs(ctx, client, namespace)
```
matthisholleville, line 21: Since both analyzers require the same codebase (findPrometheusPodConfigs, parsePromConfig, ...), shouldn't we find a way to retrieve this information once and pass it to the analyzers?

matthisholleville (changes requested):
Thank you for this awesome contribution 🤩 . I'm really excited to add this integration. Left some comments

pkg/integration/prometheus/config_analyzer.go
pintohutch: Yep - same as above.

pkg/integration/prometheus/config_analyzer.go
pintohutch: Sure - changed.

pkg/integration/prometheus/prometheus.go
pintohutch: My understanding was that `integrations` manage workloads by default, and that you'd need to specify `--no-install` explicitly when activating to avoid this.

I feel like this is a slippery slope and I'd prefer not to have the tool manage third-party workloads in the cluster.

Especially the case for Prometheus, where most clusters probably have some form of Prometheus running anyway.

The downside of this is we need to make assumptions about the characteristics of the K8s resources in-source (as done here). However, that seems preferable, and ultimately more useful than relying on users managing the tools through `k8sgpt`.

Though this may be a larger discussion. I'm sure @AlexsJones has thoughts.

pintohutch: I do agree having a bunch of `no-op` methods here isn't great 😒

pkg/integration/prometheus/relabel_analyzer.go
pintohutch: Totally - I wasn't sure how to do that....

We could put a call to `findPrometheusPodConfigs` in the integration's `Deploy()` method and we can initiate each analyzer with that state. However, then IIUC if one runs

```
k8sgpt integration activate prometheus --no-install
```

the state won't be [1] initialized...

I'm open to suggestions 😄 

[1] https://github.com/k8sgpt-ai/k8sgpt/blob/v0.3.25/pkg/integration/integration.go#L86-L90

pintohutch (commented):
Thanks @matthisholleville - pushed some changes per your comments!

pkg/integration/prometheus/prometheus.go
AlexsJones: I think in this case as long as we document on `docs.k8sgpt.ai` how this works and maybe have a message printed it's fine e.g. "Activating Prometheus integration... Found existing installation" to help connect the dots for the user.. I do agree that the install pattern is not ideal - I did that out of necessity originally, but I am actually keen to remove it from Trivy

AlexsJones:
Overall comment - great contribution and thank you for helping us figure out what we want integrations to be.  I am glad the pattern itself is holding up okay and kudos to figuring it out.

I do think there should be some guard rails if you `go run main.go integration activate prometheus` without having prometheus installed -perhaps use the `Deploy()` function to check if there is a prometheus pod in the namespace with a simple regex? At least have a warning message if we don't find something. 

In future, ( maybe kubecon ) the contributors need to talk about that wider vision of how integrations will look - I am very excited to have a Prometheus component though.

pintohutch:
Thanks @AlexsJones - added the functionality to `Deploy()` per suggestion and included some logging and comments with context.

Demo:
https://github.com/k8sgpt-ai/k8sgpt/assets/3731548/bafcec60-b772-4d5d-8288-a9511fdef3ac

pkg/integration/prometheus/prometheus.go
pintohutch: Sounds good. Filed #858 as a follow-up. I can take that on next.

pkg/integration/prometheus/relabel_analyzer.go
pintohutch: Actually - thinking about this more, it may be beneficial to rerun this with every analyze call, in case the state of the cluster changes. Called that out here [1].

[1] https://github.com/k8sgpt-ai/k8sgpt/pull/855/files#diff-666d125bbcc3df99f3456b63ade46fc4e7c0d2e27ab325c709c0e792b9143524R44

matthisholleville (approved):
LGTM

PR #904, opened 2024-02-01
test: added tests for the pkg/integration package

VaibhavMalik4187:
## 📑 Description
This commit adds new tests for the `pkg/integration` package. As a result, the code the code coverage of the package has increased from 0% to 100%

This also includes a minor adjustment in the error statements of the `Activate` and `Deactive` functions to ensure better understanding of the cause of the error.

## ✅ Checks


pkg/integration/integration.go
```diff
@@ -87,7 +87,7 @@ func (*Integration) Activate(name string, namespace string, activeFilters []stri
 
 	if !skipInstall {
 		if err := integrations[name].Deploy(namespace); err != nil {
-			return err
```
AlexsJones, line 90: This hides the inner error

arbreezy: it indeed wraps the error, this looks good to me,

but I am not sure about the value of these integration tests, they seem a bit pedantic I think

AlexsJones (approved):

PR #909, opened 2024-02-02
fix: multiple time add default backend error

### LINKED ISSUE #900: [Bug]: Able to add default backend multiple times
### Checklist

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

v0.3.26

### Kubernetes Version

v1.29.1

### Host OS and its Version

Linux

### Steps to reproduce

1. Run the `k8sgpt auth add` command without specifying the authentication backend.
2. Enter the OpenAI API secret key.
3. Run the `k8sgpt auth add` command again.

### Expected behaviour

In the command log, a warning appears indicating that a `provider with the same name already exists`.

image

### Actual behaviour

In the current command for the default authentication backend (OpenAI), we can add the authentication backend multiple times, which is not possible with other backends.

image

### Additional Information

_No response_

amitamrutiya2210:
Closes #900 

## 📑 Description
In the previous command for the default authentication backend (OpenAI), we can add the authentication backend multiple times, which is not possible with other backends. Now, in this pull request I added one condition that prevent of this behaviour and give warning message `Provider with same name already exists.`.

## ✅ Checks

## ℹ Additional Information
Any suggestions you have are appreciated and helpful.

### DIFF
--- a/cmd/auth/add.go
+++ b/cmd/auth/add.go
@@ -47,21 +47,11 @@ var addCmd = &cobra.Command{
 		}
 	},
 	Run: func(cmd *cobra.Command, args []string) {
-
 		// get ai configuration
 		err := viper.UnmarshalKey("ai", &configAI)
 		if err != nil {
 			color.Red("Error: %v", err)
 			os.Exit(1)
-		}
-
-		// search for provider with same name
-		providerIndex := -1
-		for i, provider := range configAI.Providers {
-			if backend == provider.Name {
-				providerIndex = i
-				break
-			}
 		}
 
 		validBackend := func(validBackends []string, backend string) bool {
@@ -107,6 +97,15 @@ var addCmd = &cobra.Command{
 				os.Exit(1)
 			}
 			password = strings.TrimSpace(string(bytePassword))
+		}
+
+		// search for provider with same name
+		providerIndex := -1
+		for i, provider := range configAI.Providers {
+			if backend == provider.Name || defaultBackend == provider.Name {
+				providerIndex = i
+				break
+			}
 		}
 
 		// create new provider object


cmd/auth/add.go
```diff
@@ -58,7 +58,7 @@ var addCmd = &cobra.Command{
 		// search for provider with same name
 		providerIndex := -1
 		for i, provider := range configAI.Providers {
-			if backend == provider.Name {
+			if backend == provider.Name || defaultBackend == provider.Name{
```
VaibhavMalik4187, line 61: In my opinion, instead of adding another check we should restructure the function to validate the input upfront. So the general flow of the updated function should go like:

```
1. Get ai configuration
2. Validate input values upfront and set default values if the inputs are empty
3. Find existing provider index
4. Instantiate the newProvider object.
5. Add the provider based on the value of providerIndex.
```

amitamrutiya2210: Yeah, that's also a great approach. If @AlexsJones wants to fix this issue in that way, I'll go with that approach.

AlexsJones: I agree with @VaibhavMalik4187 lets bias towards readability/maintainability here

amitamrutiya2210: Okay I will make changes and update this PR.

amitamrutiya2210:
@AlexsJones, I've made changes as per your suggestion. Please go ahead and review this PR.

cmd/auth/add.go
```diff
@@ -109,6 +99,15 @@ var addCmd = &cobra.Command{
 			password = strings.TrimSpace(string(bytePassword))
 		}
 
+		// search for provider with same name
+		providerIndex := -1
+		for i, provider := range configAI.Providers {
+			if backend == provider.Name || defaultBackend == provider.Name {
```
VaibhavMalik4187, line 105: Hi @amitamrutiya2210, I was wondering if `defaultBackend == provider.Name` is still needed? I'd suggest taking a look at https://github.com/k8sgpt-ai/k8sgpt/pull/929/files#diff-1f69cd9b70b884bda8de404da6c720e50b8aa29af0ba3365563a44f532ead7a6 for clearer understanding.

amitamrutiya2210: Yes, this is necessary until your PR is merged. In your PR, this problem is solved in a different manner with a lot of code changes. So, until your PR is not merged, this solution remains valid.

Issue #911, opened 2024-02-03
[Bug]: Able to remove default backend multiple times


VaibhavMalik4187:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### Steps to reproduce

1. Run `./bin/k8sgpt auth remove -b openai`.
2. Run `./bin/k8sgpt auth list`.

### Expected behaviour

1. Run `./bin/k8sgpt auth remove -b openai`.
Output:
```
openai deleted from the AI backend provider list
```
2. Run `./bin/k8sgpt auth list`.
Output:
```
Default: 
> openai
Active: 
> localai
Unused: 
> azureopenai
> cohere
> amazonbedrock
> amazonsagemaker
> google
> noopai
> huggingface
```

### Actual behaviour

1. Run `./bin/k8sgpt auth remove -b openai`.
Output:
```
openai deleted from the AI backend provider list
```
2. Run `./bin/k8sgpt auth list`.
Output:
```
Default: 
> openai
Active: 
> openai
> localai
Unused: 
> azureopenai
> cohere
> amazonbedrock
> amazonsagemaker
> google
> noopai
> huggingface
```

AlexsJones:
Closing as stale. No activity in 6+ months and no repro on current k8sgpt v0.4.36 / operator v0.2.27. Please reopen with a current-version repro if this is still happening.

Issue #905, opened 2024-02-02
[Bug]: Auth update command not working


amitamrutiya2210:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

v0.3.26

### Kubernetes Version

v1.29.1

### Host OS and its Version

Linux

### Steps to reproduce

1. Add auth provider by `k8sgpt auth add `
2. Update auth provider by `k8sgpt auth update [flags]`

### Expected behaviour

The `k8sgpt auth update` command should be executed in accordance with the update command's description.

### Actual behaviour

If the `k8sgpt auth update` command is used as per the description, it will display the message: "Error: accepts 1 arg(s), received 0."

![image](https://github.com/k8sgpt-ai/k8sgpt/assets/91112485/218fbf1c-453a-424b-9b89-8ded867c5381)

### Additional Information

I believe this issue is related to an incorrect description of the auth update command. I can resolve this problem if it is deemed valid.

AlexsJones:
Closing as stale. No activity in 6+ months and no repro on current k8sgpt v0.4.36 / operator v0.2.27. Please reopen with a current-version repro if this is still happening.

Issue #843, opened 2024-01-05
Feature: CLI backend configuration overhaul


bwplotka:
CLI provider configuration serves the purpose well, but it could be improved:

1. Provider configuration is done via auth command, even though, some providers does not require explicit auth in configuration (should be maybe `backend` command?)
2. You can’t have multiple backends configured for the same provider (e.g. for different models, passwords, endpoints, configurations) with unique names. To improve this some configuration change might be needed, but I saw some migration code already, maybe it can be done in smooth way.

AlexsJones:
I think we could certainly improve the way the providers/auth mapping is done and support *:1 auth:provider.

VaibhavMalik4187:
@AlexsJones I've been exploring the cmd package for a while now. I'd like to give this a try.

AlexsJones:
Sure thing, no permission needed

VaibhavMalik4187:
> Sure thing, no permission needed

Very well, I'll open a PR soon.

VaibhavMalik4187:
Hey @AlexsJones @bwplotka, I was wondering if modifying the `AIProvider` struct to something like:
```go
type AIProvider struct {
    Backend string
    Config []AIConfig
    DefaultConfig int
}

type AIConfig struct {
	Name           string  `mapstructure:"name"`
	Model          string  `mapstructure:"model"`
	Password       string  `mapstructure:"password" yaml:"password,omitempty"`
	BaseURL        string  `mapstructure:"baseurl" yaml:"baseurl,omitempty"`
	EndpointName   string  `mapstructure:"endpointname" yaml:"endpointname,omitempty"`
	Engine         string  `mapstructure:"engine" yaml:"engine,omitempty"`
	Temperature    float32 `mapstructure:"temperature" yaml:"temperature,omitempty"`
	ProviderRegion string  `mapstructure:"providerregion" yaml:"providerregion,omitempty"`
	TopP           float32 `mapstructure:"topp" yaml:"topp,omitempty"`
	MaxTokens      int     `mapstructure:"maxtokens" yaml:"maxtokens,omitempty"`
}
```
would do the trick?

And then modifying the CLI commands to ask users for the backend and a configuration name whenever they add a configuration. Maybe we can add a `--config-name` flag for the update and remove commands and if this flag is not set, then the operation will be performed on the default config for that particular backend.

AlexsJones:
[...]

PR #929, opened 2024-02-08
feat: support many:1 auth:provider mapping

VaibhavMalik4187:
## 📑 Description
Added the ability to add multiple configurations for the same backend
provider.

One to one auth:provider mapping had many bugs and users were requesting
many to one auth:provider mapping functionality for more flexibility
with backend provider configurations.

This commit also includes the addition of an optional `--config-name`
flag to allow users to specify the config-name while configuring backend
AI providers.

Changes exclusive to individual subcommands are:
1. default
  - `--config-name` flag has been added to allow users to update the 
    default configuration for a backend AI provider.
  - Setting the `--config-name` flag will only update the default
    configuration for the specified backend, it will not update the 
    `configAI.DefaultProvider`.
  - To update the `configAI.DefaultProvider`, user must leave the 
    `--config-name` flag unset.

2. list
  - The list subcommand will now respect the value of `userInput` set 
    when the user is asked to show password or not.
  - The command's output has been modified to display the backend
    providers list along with their configuration names.
  - Default configs for each backend AI provider are marked with the 
    `(Default Config)` string in bright yellow color.

3. remove
  - Users can specify either single or multiple backends when
    removing the configurations.
  - If the `--config-name` flag is not set, the "default" configuration
    for the specified backend(s) will be removed(if it is present).
  - Setting the `--config-name` flag will specify the configuration name
    to delete for each of the specified backends.

4. Update
  - Now the user can only specify a single backend provider at a time.
    This has been done because updating multiple backend configs with
    same values of parameters doesn't make any sense.
  - Removal of the `Args` function because update subcommand can have
    multiple args to set different parameters for a backend config.
  - Users can also update the configuration name for an already existing
    config.

Addresses:
- https://github.com/k8sgpt-ai/k8sgpt/issues/936
- https://github.com/k8sgpt-ai/k8sgpt/issues/911
- https://github.com/k8sgpt-ai/k8sgpt/issues/905
- https://github.com/k8sgpt-ai/k8sgpt/issues/900
- https://github.com/k8sgpt-ai/k8sgpt/issues/843

## ✅ Checks

### DIFF
--- a/cmd/analyze/analyze.go
+++ b/cmd/analyze/analyze.go
@@ -28,6 +28,7 @@ import (
 var (
 	explain         bool
 	backend         string
+	configName      string
 	output          string
 	filters         []string
 	language        string
@@ -51,6 +52,7 @@ var AnalyzeCmd = &cobra.Command{
 		// Create analysis configuration first.
 		config, err := analysis.NewAnalysis(
 			backend,
+			configName,
 			language,
 			filters,
 			namespace,
@@ -126,6 +128,8 @@ func init() {
 	AnalyzeCmd.Flags().BoolVarP(&explain, "explain", "e", false, "Explain the problem to me")
 	// add flag for backend
 	AnalyzeCmd.Flags().StringVarP(&backend, "backend", "b", "", "Backend AI provider")
+	// add flag for config-name
+	AnalyzeCmd.Flags().StringVarP(&configName, "config-name", "", "", "Name of the AI provider config to use")
 	// output as json
 	AnalyzeCmd.Flags().StringVarP(&output, "output", "o", "text", "Output format (text, json)")
 	// add language options for output
--- a/cmd/auth/add.go
+++ b/cmd/auth/add.go
@@ -28,8 +28,141 @@ import (
 
 const (
 	defaultBackend = "openai"
+	defaultConfig  = "default"
 	defaultModel   = "gpt-3.5-turbo"
 )
+
+func runAddCommand(cmd *cobra.Command, args []string) {
+	// 1. Get ai configuration
+	err := viper.UnmarshalKey("ai", &configAI)
+	if err != nil {
+		color.Red("Error: %v", err)
+		os.Exit(1)
+	}
+
+	// 2. Validate input values upfront and set default values if the inputs are empty
+	// check if backend is not empty and a valid value
+	validBackend := func(validBackends []string, backend string) bool {
+		for _, b := range validBackends {
+			if b == backend {
+				return true
+			}
+		}
+		return false
+	}
+
+	if backend == "" {
+		// Set the default value of the backend provider
+		color.Yellow(fmt.Sprintf("Warning: backend input is empty, will use the default value: %s", defaultBackend))
+		backend = defaultBackend
+	} else {
+		// Check if the given provider is valid or not.
+		if !validBackend(ai.Backends, backend) {
+			color.Red("Error: Backend AI accepted values are '%v'", strings.Join(ai.Backends, ", "))
+			os.Exit(1)
+		}
+	}
+
+	// Set the value of config-name if it is not provided by the user.
+	if configName == "" {
+		color.Yellow(fmt.Sprintf("Warning: config-name input is empty, will use the default value: %s", defaultConfig))
+		configName = defaultConfig
+	}
+
+	// 3. Find existing provider index
+	// search for provider with same backend
+	providerIndex := -1
+	configIndex := -1
+	for i, provider := range configAI.Providers {
+		if backend == provider.Backend {
+			providerIndex = i
+
+			// Iterate over all the configs of this provider
+			// and check if a config with the same name already exists
+			for index, config := range provider.Configs {
+				if configName == config.Name {
+					configIndex = index
+					break
+				}
+			}
+
+			if configIndex != -1 {
+				break
+			}
+		}
+	}
+
+	// Quit if the config already exists
+	if configIndex != -1 {
+		color.Red("Provider with same config already exists.")
+		os.Exit(1)
+	}
+
+	// Handle input sanitization for config.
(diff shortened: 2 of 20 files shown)


VaibhavMalik4187:
@AlexsJones @bwplotka I've added initial code to facilitate *:1 auth:provider mapping. In this version, the AI backend configuration is stored in the following format:

```yaml
ai:
    providers:
        - backend: openai
          configs:
            - name: default
              model: gpt-3.5-turbo
              password: some-pass
              temperature: 0.7 
              topp: 0.5 
              maxtokens: 2048
            - name: New 
              model: gpt-3.5-turbo
              password: pass
              temperature: 0.7 
              topp: 0.5 
              maxtokens: 2048
          defaultconfig: 0
        - backend: localai
          configs:
            - name: New 
              model: gpt-3.5-turbo
              temperature: 0.7 
              topp: 0.5 
              maxtokens: 2048
          defaultconfig: 0
    defaultprovider: ""
kubeconfig: ""
kubecontext: ""
```

I plan to add another subcommand to the auth command to allow the user to set the default config name for each backend provider. Before I do that, I'd like to know if this is going in the right direction.

VaibhavMalik4187:
@AlexsJones I've updated the code to allow users to set the default config for an AI backend. However, I'm unsure how to proceed with the `auth remove` and `auth update` commands. These commands process a list of backends and, with many-to-one auth: provider mapping, we'll need to specify the config name as well.

How about the following approach:
1. Take input single backend and single config-name. Delete the specified config from the specified backend.
2. Provide a `--delete-all-configs` flag, allowing the user to delete multiple backends at a time.

VaibhavMalik4187:
@AlexsJones, this PR is ready for an initial review. Could you please take a look? Thanks in advance!

VaibhavMalik4187:
@matthisholleville, could you please go through this PR whenever you have a moment? Thanks!

VaibhavMalik4187:
@JuHyung-Son could you please take a look at this PR? Thanks!

VaibhavMalik4187:
@AlexsJones ...

AlexsJones:
Apologies, I missed this one @VaibhavMalik4187 

Can you show me how you'd use a different configuration for a backend?
The commands aren't clear to me

VaibhavMalik4187:
> Apologies, I missed this one @VaibhavMalik4187
> 
> Can you show me how you'd use a different configuration for a backend? The commands aren't clear to me

No problem @AlexsJones. Sure, here we go:

1. Listing the backend configurations.
image

2. Adding a new backend without using the `config-name` flag.
image
 
3. Adding a new backend with the `config-name` flag specified.
image

4. Changing the default backend without specifying the config-name
image

5. Changing the default config for an ai provider
image

I know that this is a very large change and hope these examples make things slightly easier to understand.

JuHyung-Son:
I haven't tested it, but it looks good from the screenshots. However I think we need to discuss if this change to enable multiple configs is really necessary.

VaibhavMalik4187:
> I haven't tested it, but it looks good from the screenshots. �However I think we need to discuss if this change to enable multiple configs is really necessary.

If we agree to move forward with this change, we'll also need to update the operator accordingly.

AlexsJones:
> > I haven't tested it, but it looks good from the screenshots. �However I think we need to discuss if this change to enable multiple configs is really necessary.
> 
> If we agree to move forward with this change, we'll also need to update the operator accordingly.

Would we have to, or is this more about parity?

VaibhavMalik4187:
> > > I haven't tested it, but it looks good from the screenshots. �However I think we need to discuss if this change to enable multiple configs is really necessary.
> > 
> > 
> > If we agree to move forward with this change, we'll also need to update the operator accordingly.
> 
> Would we have to, or is this more about parity?

I think it is necessary to include the `ConfigName` field in the `AnalyzeRequest` data structure (see https://github.com/k8sgpt-ai/k8sgpt/pull/929/files#diff-81b03b26debf92e0e0bcaa55bdfcfbfc72a5a9ebafa949bb80b2474e759f506a)

JuHyung-Son:
> > > > I haven't tested it, but it looks good from the screenshots. �However I think we need to discuss if this change to enable multiple configs is really necessary.
> > > 
> > > 
> > > If we agree to move forward with this change, we'll also need to update the operator accordingly.
> > 
> > 
> > Would we have to, or is this more about parity?
> 
> I think it is necessary to include the `ConfigName` field in the `AnalyzeRequest` data structure (see https://github.com/k8sgpt-ai/k8sgpt/pull/929/files#diff-81b03b26debf92e0e0bcaa55bdfcfbfc72a5a9ebafa949bb80b2474e759f506a)

Im not that sure we need to provide multiple auth keys. I think it's unnecessarily complicated and hinders the user experience a lot, and the user can just replace it with a different config via the `k8sgpt -c` option.

VaibhavMalik4187:
> > > > > I haven't tested it, but it looks good from the screenshots. �However I think we need to discuss if this change to enable multiple configs is really necessary.
> > > > 
> > > > 
> > > > If we agree to move forward with this change, we'll also need to update the operator accordingly.
> > > 
> > > 
> > > Would we have to, or is this more about parity?
> > 
> > 
> > I think it is necessary to include the `ConfigName` field in the `AnalyzeRequest` data structure (see https://github.com/k8sgpt-ai/k8sgpt/pull/929/files#diff-81b03b26debf92e0e0bcaa55bdfcfbfc72a5a9ebafa949bb80b2474e759f506a)
> 
> Im not that sure we need to provide multiple auth keys. I think it's unnecessarily complicated and hinders the user experience a lot, and the user can just replace it with a different config via the `k8sgpt -c` option.

Hmm, fair enough. In my opinion, these changes won'thinder the user experience but provide more flexible approach to setting the AI configurations. The worst part is this PR also fixes a bunch of issues other than many:1 auth provider mapping.

PR #966, opened 2024-02-18
fix: analyze command default backend bug

### LINKED ISSUE #902: [Bug]: analyze`s --backend arg error
### Checklist

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

_No response_

### Kubernetes Version

_No response_

### Host OS and its Version

_No response_

### Steps to reproduce

2 problems.

1. set default backend as none openai and give -b openai -> use default backend not openai
```
> ./bin/k8sgpt analyze -f Service -n default -c -e -b openai
   0% |                                                                                                                                                                                                                                                                                | (0/8, 0 it/hr) [0s:0s]
Error: failed while calling AI provider localai: error, status code: 404, message: %!s(<nil>)
```

2. set any default backend and run analyze with -b arg -> only use default backend not provided arg. (i think this is bug)
```
> ./bin/k8sgpt auth default -p noopai
Default provider set to noopai
> ./bin/k8sgpt analyze -f Service -n default -e -c localai
 100% |███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| (7/7, 2693 it/s)
AI Provider: noopai
```

### Expected behaviour

ai provider should be provided `--backend`  even `openai` is given.

### Actual behaviour

_No response_

### Additional Information

_No response_

VaibhavMalik4187:
Closes #902

## 📑 Description
Now, the default value of the `backend` flag for the analyze command will be an empty string. The `NewAnalysis` function has been modified to use the default backend set by the user if the backend flag is not provided and the `defaultprovider` is set in the config file. Otherwise, the backend will be set to "openai".

## ✅ Checks


VaibhavMalik4187:
@JuHyung-Son could you please go through this PR whenever you have some time? Thanks!

AlexsJones:
This looks like it's just moving the functionality elsewhere?

pkg/analysis/analysis.go
```diff
@@ -125,10 +125,14 @@ func NewAnalysis(
 
 	// Backend string will have high priority than a default provider
 	// Backend as "openai" represents the default CLI argument passed through
```
JuHyung-Son, line 127: comments need to be fixed

pkg/server/analyze.go
```diff
@@ -16,10 +16,6 @@ func (h *handler) Analyze(ctx context.Context, i *schemav1.AnalyzeRequest) (
 		i.Output = "json"
 	}
 
-	if i.Backend == "" {
```
JuHyung-Son, line 19: why does this lines are removed?

JuHyung-Son (commented):
thanks, cli backend options are working well. and i left 2 comments

pkg/analysis/analysis.go
VaibhavMalik4187: Done, thanks for pointing this out.

pkg/server/analyze.go
VaibhavMalik4187: This code was setting the empty backend variable to "openai" (Default value). However, the latest version of the `NewAnalysis` function handles an empty `backend` value in the function itself. Hence, I removed this code.

VaibhavMalik4187:
> This looks like it's just moving the functionality elsewhere?

Consider the following scenario if this fix was not present:
1. User sets the default provider to say `localai`
2. Suppose, I've also configured an "openai" backend and instead of using the default provider, I want to use `openai` to `analyze` the cluster.
3. How would I do that by specifying the `backend` flag? These conditions would enter the following if block and force set the backend to `localai`:
```
if configAI.DefaultProvider != "" && backend == "openai" {
```

However, using the fix I proposed, the user can also specify `openai` as the backend for the `analyze` command. And setting the default value of the `backend` flag as `""` helps in identifying if the user wanted to `openai` or not.

JuHyung-Son:
i think i looks good. and works well.
and im not sure this problem is related to this pr. doe `k8sgpt serve` works??

CleanShot 2024-02-19 at 11 10 47

VaibhavMalik4187:
> i think i looks good. and works well. and im not sure this problem is related to this pr. doe `k8sgpt serve` works??
> 
> CleanShot 2024-02-19 at 11 10 47

I think you hunch is right, this is a separate problem because I saw changes related to `cmd/serve/serve.go` when rebasing https://github.com/k8sgpt-ai/k8sgpt/pull/929 on main. I'll look into it and open a PR with the fix.

JuHyung-Son (approved):
sorry for delay, this looks good to me.
and cli works well

VaibhavMalik4187:
> sorry for delay, this looks good to me. and cli works well

No worries, I've rebased the PR to included the latest changes on main.

Issue #889, opened 2024-01-24
Increase Analysers' test coverage


arbreezy:
K8sGPT analysers are the first step of K8sGPT analysis whereby we scan K8s clusters and identify issues with K8s resources.
To gain more confidence every analyser should come with its own unit testing of various use cases we aim to identify.

There are a few analysers that have either limited or absent unit tests. The goal of this work is to increase code coverage of K8sGPT analysers by mocking a K8s environment. e.g [Statefulset analyser](https://github.com/k8sgpt-ai/k8sgpt/blob/main/pkg/analyzer/statefulset_test.go)

Here is a non-exhaustive list but a rather good start of enhancing or adding unit testing.

- validating webhook
- pvc
- replicaset
- events
- logs analyser
- analyzers' wrapper functions

AkashKumar7902:
Hey @arbreezy 
I am interested in working on this project. My internship with the CNCF landscape project, [keploy](https://github.com/keploy/keploy), has provided me with practical experience in enhancing test coverage. I would love to start understanding the codebase by solving few issues.

Ishani217:
Hey @arbreezy 

I was recently testing the `statefulset.go` file and running the `statefulset_test.go` gave me errors below:

```bash
# command-line-arguments [command-line-arguments.test]
./statefulset_test.go:38:25: undefined: StatefulSetAnalyzer
./statefulset_test.go:65:25: undefined: StatefulSetAnalyzer
./statefulset_test.go:130:25: undefined: StatefulSetAnalyzer
./statefulset_test.go:176:25: undefined: StatefulSetAnalyzer
FAIL    command-line-arguments [build failed]
FAIL
```

Could you please specify a procedure to run tests correctly (I mean to run the existing tests, if any)

arbreezy:
Heu @Ishani217 you can use the Makefile to run the tests, `make test`

amitamrutiya2210:
Hello everyone 👋,

I have a basic understanding of Linux, Docker, Helm, Kubernetes, Go, version control, and GitHub Actions. I participated in GSoC 2023 with CCExtractor, working on the open-source [flood-mobile](https://github.com/CCExtractor/Flood_Mobile) project. Now, I plan to join the k8sgpt project through the LFX Mentorship Program.

[...]

PR #1010, opened 2024-03-12
test: added tests for the log analyzer

VaibhavMalik4187:
## 📑 Description
* Added new tests for `LogAnalyzer` defined in the `pkg/analyzer` package. Increased the code coverage of the `log.go` file to >90%

* Added `fake logs` string to the `errorPattern` to enable the discovery of error logs during the tests.

TODO: Find a method to edit logs of the containers defined in the Kubernetes client config.

Partially addresses: https://github.com/k8sgpt-ai/k8sgpt/issues/889

## ✅ Checks

### DIFF
--- a/pkg/analyzer/log_test.go
+++ b/pkg/analyzer/log_test.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package analyzer
+
+import (
+	"context"
+	"regexp"
+	"sort"
+	"testing"
+
+	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
+	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestLogAnalyzer(t *testing.T) {
+	oldPattern := errorPattern
+	errorPattern = regexp.MustCompile(`(fake logs)`)
+	t.Cleanup(func() {
+		errorPattern = oldPattern
+	})
+
+	config := common.Analyzer{
+		Client: &kubernetes.Client{
+			Client: fake.NewSimpleClientset(
+				&v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "Pod1",
+						Namespace: "default",
+						Labels: map[string]string{
+							"Name":      "Pod1",
+							"Namespace": "default",
+						},
+					},
+					Spec: v1.PodSpec{
+						Containers: []v1.Container{
+							{
+								Name: "test-container1",
+							},
+							{
+								Name: "test-container2",
+							},
+						},
+					},
+				},
+				&v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "Pod2",
+						Namespace: "default",
+						Labels: map[string]string{
+							"Name":      "Pod1",
+							"Namespace": "default",
+						},
+					},
+				},
+				&v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "Pod3",
+						Namespace: "test-namespace",
+						Labels: map[string]string{
+							"Name":      "Pod1",
+							"Namespace": "test-namespace",
+						},
+					},
+				},
+				&v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


pkg/analyzer/log_test.go
```diff
@@ -0,0 +1,113 @@
...
+	require.NoError(t, err)
+
+	sort.Slice(results, func(i, j int) bool {
+		return results[i].Name < results[j].Name
+	})
+
+	expectations := []string{"default/Pod1/test-container1", "default/Pod1/test-container2", "default/Pod4/test-container3"}
+
+	for i, expectation := range expectations {
+		require.Equal(t, expectation, results[i].Name)
```
JuHyung-Son, line 107: what is purpose of this test?

pkg/analyzer/log_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+		for _, failure := range results[i].Error {
+			require.Equal(t, "fake logs", failure.Text)
```
JuHyung-Son, line 110: how does these fake pods have error logs??

JuHyung-Son (commented):
good work! i commented two questions

pkg/analyzer/log_test.go
VaibhavMalik4187: It is to verify that only the right pods and containers are reported in the results.

pkg/analyzer/log_test.go
VaibhavMalik4187: My had, I should've marked this PR as draft. I've added a TODO note in the description to find a method to edit the logs of the fake containers.

Currently, the logs of these containers are automatically set to "fake logs" during instantiation. This is why I've marked "fake logs" as an erroneous log.

pkg/analyzer/log_test.go
JuHyung-Son: oh i see. i missed it

pkg/analyzer/log_test.go
VaibhavMalik4187: @JuHyung-Son, I tried but couldn't find any method to modify the logs of these fake containers :(

Could you please share any ideas for creating fake logs?

JuHyung-Son: yeah.. i think there`s no way to do it.
i think it would be better to define fake logs or `results[i].Error` corresponding to each pod. think it as mock like `fake.NewSimpleClientset` create fake pods.

and remove 'fake logs' in this line. it's only for testing.
`errorPattern = regexp.MustCompile('(error|exception|fail|fake logs)')`

VaibhavMalik4187: If I remove `fake logs` from `errorPattern = regexp.MustCompile('(error|exception|fail|fake logs)')`, then there won't be anything in the results at all.

JuHyung-Son: yeah. i confused before. If we define `Error`, test is meaningless.
I'm not sure what to do about this situation. Maybe it's best to leave now.

@AlexsJones how do you think?

AlexsJones: I think this was always going to be a difficult one to test.  I am not sure, will need to think about it

arbreezy: maybe we can try doing something like this and stream fake logs

```
import (
	"bytes"
	"context"
	"io"
	"testing"

	corev1 "k8s.io/api/core/v1"
	cgtesting "k8s.io/client-go/testing"
)
fp := FakePods{
		Fake: &FakeCoreV1{Fake: &cgtesting.Fake{}},
		ns:   "default",
	}
	req := fp.GetLogs("foo", &corev1.PodLogOptions{})
	body, err := req.Stream(context.Background())
	if err != nil {
		t.Fatal("Stream pod logs:", err)
	}
	var buf bytes.Buffer
	n, err := io.Copy(&buf, body)
	if err != nil {
		t.Fatal("Read pod logs:", err)
	}
	if n == 0 {
		t.Fatal("Empty log")
	}
	err = body.Close()
	if err != nil {
		t.Fatal("Close response body:", err)
	}
}```
an [example](https://github.com/kubernetes/kubernetes/pull/91485/files)

VaibhavMalik4187: That looks very helpful, thanks a lot @arbreezy! I'll update this PR soon.

arbreezy: nice overview, let me investigate a little bit more before we concede :)

VaibhavMalik4187: Sure thing.

VaibhavMalik4187: @arbreezy, any updates on this?

arbreezy: My approach ended up, being really hacky and to be fair not worth it. I think if we want to test the regex we can do it independetly of mocking logs in a pod.

VaibhavMalik4187: Agreed, we could always test regex separately. As a last resort, do you think that it would be a good idea to raise a feature request to allow passing a string to initialize the pod logs?

AlexsJones: Let's go with what we have for now?

AlexsJones (approved):
I am okay with this as a first pass

pkg/analyzer/log_test.go
arbreezy: Yes, makes sense.
I was OOO sorry for the radio silence

VaibhavMalik4187:
> I am okay with this as a first pass

Thanks, I've rebased this branch to include the latest changes.

pkg/analyzer/log_test.go
VaibhavMalik4187: No problem @arbreezy

PR #1019, opened 2024-03-16
test: added missing tests for the CronJob analyzer

VaibhavMalik4187:
## 📑 Description
- Fixed a small bug where pre-analysis was incorrectly appended to the results every time at the end of the for loop. This caused the result for a single cronjob failure to be appended multiple times in the final results.

- Added missing test cases to ensure proper testing of the CronJob analyzer. The addition of these missing test cases has increased the code coverage of this analyzer to over 96%.

Partially Addresses: https://github.com/k8sgpt-ai/k8sgpt/issues/889

## ✅ Checks

### DIFF
--- a/pkg/analyzer/cronjob_test.go
+++ b/pkg/analyzer/cronjob_test.go
@@ -15,219 +15,165 @@ package analyzer
 
 import (
 	"context"
+	"sort"
 	"testing"
 
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
-	"github.com/magiconair/properties/assert"
+	"github.com/stretchr/testify/require"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
 )
 
-func TestCronJobSuccess(t *testing.T) {
-	clientset := fake.NewSimpleClientset(&batchv1.CronJob{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "example-cronjob",
-			Namespace: "default",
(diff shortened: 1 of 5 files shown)


pkg/analyzer/cronjob_test.go
```diff
@@ -15,219 +15,165 @@ package analyzer
...
 
-	config := common.Analyzer{
-		Client: &kubernetes.Client{
-			Client: clientset,
-		},
-		Context:   context.Background(),
-		Namespace: "default",
-	}
+	startingDeadline := new(int64)
+	*startingDeadline = -7
```
JuHyung-Son, line 35: It would be nice to be able to tell if a variable is for a failure case or not just by its name.

JuHyung-Son (commented):
nice catch.

just one question.
i guess comparing failureText can occur unexpected test fail in the future even code is right.
how do you think?

VaibhavMalik4187:
> nice catch.
> 
> just one question. i guess comparing failureText can occur unexpected test fail in the future even code is right. how do you think?

Thanks,

Great question. I think you're right, there's a possibility of the failure text changing in the future and when that happens the test will break even though it shouldn't. 

I've compared failureText in tests for others analyzers as well. Looks like now would be a great time to update them as well. Thought?

pkg/analyzer/cronjob_test.go
VaibhavMalik4187: Thanks for the suggestion! Cleaner code is always helpful for everyone.

JuHyung-Son:
> > nice catch.
> > just one question. i guess comparing failureText can occur unexpected test fail in the future even code is right. how do you think?
> 
> Thanks,
> 
> Great question. I think you're right, there's a possibility of the failure text changing in the future and when that happens the test will break even though it shouldn't.
> 
> I've compared failureText in tests for others analyzers as well. Looks like now would be a great time to update them as well. Thought?

yes. it needs to be. but not urgent. feel free to do that.

PR #1042, opened 2024-03-26
refactor: replace getEnv with cmp.or & make fmt

izturn:
## 📑 Description

replace getEnv with cmp.Or  & make fmt

## ✅ Checks

## ℹ Additional Information

Signed-off-by: Gang Liu [gang.liu@daocloud.io](mailto:gang.liu@daocloud.io)

### DIFF
--- a/pkg/integration/trivy/trivy.go
+++ b/pkg/integration/trivy/trivy.go
@@ -14,6 +14,7 @@ limitations under the License.
 package trivy
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/integration/trivy/trivy.go
```diff
@@ -30,25 +31,17 @@ import (
 )
 
 var (
-	Repo          = getEnv("TRIVY_REPO", "https://aquasecurity.github.io/helm-charts/")
-	Version       = getEnv("TRIVY_VERSION", "0.13.0")
-	ChartName     = getEnv("TRIVY_CHART_NAME", "trivy-operator")
-	RepoShortName = getEnv("TRIVY_REPO_SHORT_NAME", "aqua")
-	ReleaseName   = getEnv("TRIVY_RELEASE_NAME", "trivy-operator-k8sgpt")
+	Repo          = cmp.Or("TRIVY_REPO", "https://aquasecurity.github.io/helm-charts/")
```
arbreezy, line 34: I think that doesn't look right,
 the desired behavior is to see if an env variable is populated.

cmp.Or [function](https://pkg.go.dev/cmp#Or) in this implementation will make Repo's value to `TRIVY_REPO` string

I also don't see the value of using cmp to replace a small function

arbreezy (changes requested):

pkg/integration/trivy/trivy.go
izturn: got it

PR #1053, opened 2024-04-04
fix: set topP from config

### LINKED ISSUE #1052: [BUG]: topP is not set from the config yaml of the `openai` provider
### Checklist

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

v0.3.29

### Kubernetes Version

v1.27

### Host OS and its Version

MacOS

### Steps to reproduce

1. Create an `openai` provider with a topp different than `1`
2. If you log the request, the parameter is hardcoded to 1 directly in the code
    see line: https://github.com/k8sgpt-ai/k8sgpt/blob/6b38a56afbdaa8e0d8f025088a52d3022673ef9d/pkg/ai/openai.go#L40

### Expected behaviour

As `topP` is a valid parameter from the openAI spec: https://platform.openai.com/docs/api-reference/chat/create#chat-create-top_p; a user should be able to set this parameter. 

In addition, there are many `openai` compatible backends that could take advantage of a user defined `topP`.

### Actual behaviour

`topP` should be set from the config file. 

### Additional Information

_No response_

muscionig:
Closes #1052 

## 📑 Description
This PR allows for setting topP from the config yaml for the `openai` provider.

## ✅ Checks

## ℹ Additional Information
I took the liberty of opening both the issue and the PR fixing it, please let me know if I should follow any other process.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -27,6 +27,7 @@ import (
 
 const (
 	defaultTemperature float32 = 0.7
+	defaultTopP        float32 = 1.0
 )
 
 var (
@@ -67,6 +68,22 @@ var ServeCmd = &cobra.Command{
 				}
 				return float32(temperature)
 			}
+			topP := func() float32 {
+				env := os.Getenv("K8SGPT_TOP_P")
+				if env == "" {
+					return defaultTopP
+				}
+				topP, err := strconv.ParseFloat(env, 32)
+				if err != nil {
+					color.Red("Unable to convert topP value: %v", err)
+					os.Exit(1)
+				}
+				if topP > 1.0 || topP < 0.0 {
+					color.Red("Error: topP ranges from 0 to 1.")
+					os.Exit(1)
+				}
+				return float32(topP)
+			}
 			// Check for env injection
 			backend = os.Getenv("K8SGPT_BACKEND")
 			password := os.Getenv("K8SGPT_PASSWORD")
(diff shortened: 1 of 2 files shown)


AlexsJones:
Thanks for opening a PR, I think for this to work though, you'll need to update the command/auth with a new flag for topP and link it into the configuration object

muscionig:
@AlexsJones thanks for the comment. That fag is already there in the auth command and if `--topp` is passed it is properly saved in the config yaml. However, because of the `const` value in the `openai` code, that flag is never set properly, which is the reason why I opened this as a _fix_ and not an _enhancement_.

With the edits in this PR, that parameter is properly set from the config if the user provides it.

arbreezy:
> @AlexsJones thanks for the comment. That flag is already there in the auth command and if `--topp` is passed it is properly saved in the config yaml. However, because of the `const` value in the `openai` code, that flag is never set properly, which is the reason why I opened this as a _fix_ and not an _enhancement_.
> 
> With the edits in this PR, that parameter is properly set from the config if the user provides it.

You are right, probably we never removed the constant when introducing the topP as a cli argument.

Could you please, also add it as an env variable in server's cmd [file](https://github.com/k8sgpt-ai/k8sgpt/blob/main/cmd/serve/serve.go#L70) so we can use it in the operator?

muscionig:
@arbreezy thanks for reviewing this. I took care of the format and I have also enabled the setting of `topP` in the operator, I followed the way `temperature` is set. For the default value, I did not alter the previous behavior (which is also the recommended use from openai and other providers).

arbreezy:
> @arbreezy thanks for reviewing this. I took care of the format and I have also enabled the setting of `topP` in the operator, I followed the way `temperature` is set. For the default value, I did not alter the previous behavior (which is also the recommended use from openai and other providers).

It looks good, after resolving the conflicts I'll approve 👍

arbreezy (approved):

PR #1068, opened 2024-04-17
fix: invalid ParentObj in output

### LINKED ISSUE #1067: [Bug]:  invalid parent object in output list
### Checklist

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

_No response_

### Kubernetes Version

_No response_

### Host OS and its Version

_No response_

### Steps to reproduce

`k8sgpt analyze`

### Expected behaviour

```
1 default/tt8(tt8)
- Error: Service has no endpoints, expected label k=v
```

for those objects without parentObject, the blanket will show nothing as below

```
1 default/tt8()
- Error: Service has no endpoints, expected label k=v
```

### Actual behaviour

_No response_

### Additional Information

_No response_

panpan0000:
The `GetParent()` always return `false` and the invoker never respect its return value.

So add the boolean return value (I believe it means "found" )

Closes #1067

## ✅ Checks

## ℹ Additional Information

Previous
image
Now
image


arbreezy:
Nice catch @panpan0000

pkg/util/util.go
```diff
@@ -114,11 +114,11 @@ func GetParent(client *kubernetes.Client, meta metav1.ObjectMeta) (string, bool)
 				if vw.OwnerReferences != nil {
 					return GetParent(client, vw.ObjectMeta)
 				}
-				return "ValidatingWebhook/" + vw.Name, false
+				return "ValidatingWebhook/" + vw.Name, true
 			}
 		}
 	}
-	return meta.Name, false
```
panpan0000, line 121: I'm not sure about this line . can you help

panpan0000: @AlexsJones  , can you help ?

arbreezy: sorry @panpan0000 just saw it,
yes, I think an empty string makes more sense, the meta.Name doesn't give any more information in the output
 e.g
`return "", false`

arbreezy (changes requested):

pkg/util/util.go
AlexsJones: This should return false

panpan0000: thank you both, It's fixed.

arbreezy (approved):

arbreezy:
@panpan0000 we need to fix this test as well , https://github.com/k8sgpt-ai/k8sgpt/blob/main/pkg/util/util_test.go#L85

panpan0000:
> @panpan0000 we need to fix this test as well , https://github.com/k8sgpt-ai/k8sgpt/blob/main/pkg/util/util_test.go#L85

Thank you for reminder. @arbreezy , fixed as well.

PR #1095, opened 2024-05-04
Fix prompt written answers in custom language

winggundamth:
## 📑 Description
Changing language in the answer with default_prompt is not correct.

## ✅ Checks

### DIFF
--- a/pkg/ai/prompts.go
+++ b/pkg/ai/prompts.go
@@ -1,7 +1,8 @@
 package ai
 
 const (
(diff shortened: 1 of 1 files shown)


AlexsJones:
Thanks @winggundamth I am happy to change this, but I would say this is very dependent on the LLM being used. What was this tested with?

JuHyung-Son:
does current language prompt not works on that models?

PR #1101, opened 2024-05-08
feat: added support for A21 and Amazon Titan models via bedrock api

### LINKED ISSUE #1076: [Question]: Integration with Amazon bedrock issue
### Checklist

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

0.3.29

### Kubernetes Version

v1.29

### Host OS and its Version

MacOS

### Steps to reproduce

1. I try to add aws bedrock as provider, follow this wiki to setup: [https://docs.k8sgpt.ai/reference/providers/backend/#amazon-bedrock](https://docs.k8sgpt.ai/reference/providers/backend/#amazon-bedrock)
  command is : k8sgpt auth add --backend amazonbedrock --model ai21.j2-ultra-v1 --providerRegion us-east-1
2. I setup ENV, add permission in aws IAM, and used a command (aws bedrock list-foundation-models) to test, I can get model list.
3. Error: but, when I use 'k8sgpt analyze -e -b amazonbedrock' to run, I got a error: `Error: failed while calling AI provider amazonbedrock: AccessDeniedException: You don't have access to the model with the specified model ID.`.
4. Based on [source code ](https://github.com/k8sgpt-ai/k8sgpt/blob/main/pkg/ai/amazonbedrock.go), seems like k8sgpt for now only support Claude model?

### Expected behaviour

integration with bedrock success.

### Actual behaviour

Error: failed while calling AI provider amazonbedrock: AccessDeniedException: You don't have access to the model with the specified model ID.

### Additional Information

_No response_

awsyshah:
Closes #1076 

## 📑 Description
Included support for A21 & Amazon Titan Model through the bedrock api, using a case structure to enable future flexibility to continue adding other models available. As Bedrock supports quite a few models not adding them all in a single merge.

## ✅ Checks


pkg/ai/amazonbedrock.go
```diff
@@ -116,13 +122,37 @@ func (a *AmazonBedRockClient) Configure(config IAIConfig) error {
...
-		"top_p":                0.9,
-	}
+	// Prepare the input data for the model invocation based on the model
+	var request map[string]interface{}
+
+	switch a.model {
+    case ModelAnthropicClaudeV2, ModelAnthropicClaudeV1, ModelAnthropicClaudeInstantV1:
+        request = map[string]interface{}{
+            "prompt":               fmt.Sprintf("\n\nHuman: %s  \n\nAssistant:", prompt),
+            "max_tokens_to_sample": 1024,
```
AlexsJones, line 132: Can we please use a const throughout rather than `1024` directly?

pkg/ai/amazonbedrock.go
```diff
@@ -116,13 +122,37 @@ func (a *AmazonBedRockClient) Configure(config IAIConfig) error {
...
-		"temperature":          a.temperature,
-		"top_p":                0.9,
-	}
+	// Prepare the input data for the model invocation based on the model
+	var request map[string]interface{}
+
+	switch a.model {
+    case ModelAnthropicClaudeV2, ModelAnthropicClaudeV1, ModelAnthropicClaudeInstantV1:
+        request = map[string]interface{}{
+            "prompt":               fmt.Sprintf("\n\nHuman: %s  \n\nAssistant:", prompt),
```
AlexsJones, line 131: Can we please const out this string if we're repeating multiple times?

AlexsJones (changes requested):

awsyshah:
@AlexsJones did you get a chance to review this for the changes and if it is ok to merge now ?

AlexsJones:
I need to test it first, will try to get around to it ASAP

awsyshah:
> I need to test it first, will try to get around to it ASAP

@AlexsJones just adding some screenshots of the runs if that helps.

image

image

image

lenaxia:
If you're looking for a workaround, would suggest using LiteLLM to proxy Bedrock

AlexsJones (approved):

pkg/ai/amazonbedrock.go
```diff
@@ -142,15 +167,56 @@ func (a *AmazonBedRockClient) GetCompletion(ctx context.Context, prompt string)
...
+			Stop_reason string `json:"stop_reason"`
+		}
+		output := &InvokeModelResponseBody{}
+		err = json.Unmarshal(resp.Body, output)
+		if err != nil {
+			return "", err
+		}
+        return output.Completion, nil
+    case ModelA21J2UltraV1, ModelA21J2JumboInstruct:
+		type Data struct {
```
JuHyung-Son, line 185: Is it good to define these types in case??
I suggest to put these out of case.

AlexsJones:
Can you respond to comments and we will merge this next week @awsyshah ?

awsyshah:
made the key suggest improvement of moving some of the constants to config. The other vars are as per the API or not need outside of the case and hence declared within it. Please review.

JuHyung-Son:
comments are not fixed

awsyshah:
> comments are not fixed

Hi @JuHyung-Son ,

Sorry for the trouble there, can I request which are the specific ones you want changed.

The first action of moving the constants like maxTokens to config is done right ? or is my understanding / implementation incorrect ?

For Comment 2 , the stop_reason , that call used is as per the documentation from :https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html 

For Comment 3:  The type you have commented on is only used once and that is if it is a a21 model so its defined within the area so as its only ever used there. The rest of the code has not use of it , is it still better to just define it up top even is only ever to be used if the model is one of the A21 one and may not be used otherwise ?

Let me know and I will make the changes so we can have this merged.

Thanks

pkg/ai/amazonbedrock.go
JuHyung-Son: My suggestion is put type on out of the functions. I think this will break the readability of the code and will have to be fixed anyway if any changes are made in the future.

JuHyung-Son (approved):
i got it.
LGTM!

AlexsJones (approved):

PR #1134, opened 2024-06-08
fix: typo in prometheus.go

ronaldpetty:
- fix: typo in prometheus.go


arbreezy:
Nice catch, we just follow [conventional](https://www.conventionalcommits.org/en/v1.0.0/) commits that's the reason pipeline is complaining.
Will approve it once the commit message is updated !

ronaldpetty:
I think I got it. If not please tell me and I'll look again.

Issue #1166, opened 2024-06-25
k3s node Condition `EtcdIsVoter` Status: True is Normal Considtion but reported as an error.


reefland:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

k8sgpt: 0.3.37 (3f80bba)

### Kubernetes Version

k3s v1.29.5+k3s1

### Host OS and its Version

Ubuntu 22.04.4

### Steps to reproduce

Standard k3s installation. Control-Plane nodes include a status if they are a voting etcd member or a learning member.
```
$ k describe node k3s01
...
Conditions:
  Type             Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----             ------  -----------------                 ------------------                ------                       -------
  MemoryPressure   False   Mon, 24 Jun 2024 21:06:18 -0400   Thu, 06 Jun 2024 13:33:15 -0400   KubeletHasSufficientMemory   kubelet has sufficient memory available
  DiskPressure     False   Mon, 24 Jun 2024 21:06:18 -0400   Thu, 06 Jun 2024 13:33:15 -0400   KubeletHasNoDiskPressure     kubelet has no disk pressure
  PIDPressure      False   Mon, 24 Jun 2024 21:06:18 -0400   Thu, 06 Jun 2024 13:33:15 -0400   KubeletHasSufficientPID      kubelet has sufficient PID available
  Ready            True    Mon, 24 Jun 2024 21:06:18 -0400   Fri, 14 Jun 2024 14:16:23 -0400   KubeletReady                 kubelet is posting ready status. AppArmor enabled
  EtcdIsVoter      True    Mon, 24 Jun 2024 21:06:38 -0400   Fri, 21 Jun 2024 20:02:30 -0400   MemberNotLearner             Node is a voting member of the etcd cluster
```

```
$ k8sgpt analyze
AI Provider: AI not used; --explain not set

0: Node k3s01()
- Error: k3s01 has condition of type EtcdIsVoter, reason MemberNotLearner: Node is a voting member of the etcd cluster

1: Node k3s02()
- Error: k3s02 has condition of type EtcdIsVoter, reason MemberNotLearner: Node is a voting member of the etcd cluster

2: Node k3s03()
[...]

PR #1167, opened 2024-06-25
feat: skip k3s node type `EtcdIsVoter`

yankay:
Closes  https://github.com/k8sgpt-ai/k8sgpt/issues/1166

## 📑 Description

Skip k3s node Condition EtcdIsVoter Status reported as an error.

## ✅ Checks


JuHyung-Son (commented):
does this error is only specific to k3s?

yankay:
> does this error is only specific to k3s?

Yes, it's only for k3s

JuHyung-Son (commented):
How about define `ignore []string` and put this error in it.
Then, case if NodeConditionType is in ignore, break.

AlexsJones:
I am inclined to agree that if this unblocks errors in k3s then we should add it. 

Anyone else @k8sgpt-ai/k8sgpt-maintainers  ?

arbreezy:
no objections from my side, looks safe :) @AlexsJones

Issue #1180, opened 2024-07-10
[Bug]: k8sgpt serve panic: runtime error: invalid memory address or nil pointer dereference


shturec:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

0.3.37

### Kubernetes Version

n/a

### Host OS and its Version

Mac

### Steps to reproduce

1. With a default configuration (/Users//Library/Application Support/k8sgpt/k8sgpt.yaml) including multiple specific backends, neither of which is OpenAI
2. Run k8sgpt serve

### Expected behaviour

Descriptive error for missing --backend flag with non-default (openai) backend configurations.

### Actual behaviour

k8sgpt serve
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x102e297e0]

goroutine 1 [running]:
github.com/k8sgpt-ai/k8sgpt/cmd/serve.init.func1(0x140001dce00?, {0x102e2d786?, 0x4?, 0x102e2d78a?})
github.com/k8sgpt-ai/k8sgpt/cmd/serve/serve.go:152 +0x500
github.com/spf13/cobra.(*Command).execute(0x105b5b380, {0x105d9d388, 0x0, 0x0})
github.com/spf13/cobra@v1.8.0/command.go:987 +0x828
github.com/spf13/cobra.(*Command).ExecuteC(0x105b57160)
github.com/spf13/cobra@v1.8.0/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(...)
github.com/spf13/cobra@v1.8.0/command.go:1039
github.com/k8sgpt-ai/k8sgpt/cmd.Execute({0x1036baa18?, 0x0?}, {0x1036bde88?, 0x100c567c0?}, {0x1036bdf70?, 0x140000021c0?})
github.com/k8sgpt-ai/k8sgpt/cmd/root.go:59 +0xa0
main.main()
github.com/k8sgpt-ai/k8sgpt/main.go:25 +0x4c

### Additional Information

configuration:

ai:
    providers:
        - name: localai
          model: llama3
          baseurl: http://localhost:11434/v1
          temperature: 0.7
          topp: 0.5
          topk: 50
          maxtokens: 2048
[...]

PR #1185, opened 2024-07-11
fix:k8sgpt serve panic- runtime error-invalid memory address or nil pointer dereference 

kameriso-zga:
bug:https://github.com/k8sgpt-ai/k8sgpt/issues/1180

### DIFF
--- a/pkg/server/analyze.go
+++ b/pkg/server/analyze.go
@@ -2,7 +2,7 @@ package server
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


arbreezy:
@kameriso-zga thanks for the PR, lgtm.
Can you sort out the conflicts and the DCO signing part along with a semantic PR message.

Happy to help if you get blocked

AlexsJones (approved):

PR #1201, opened 2024-07-16
feat: add label selector

### LINKED ISSUE #993: add label selector on analyzer cmd
add label selector like `kubectl get pod -l app=test`

cmd document will be same as kubectl description

```
-L, --label-columns=[]:
	Accepts a comma separated list of labels that are going to be presented as columns. Names are case-sensitive.
	You can also use multiple flag options like -L label1 -L label2...
```

JuHyung-Son:
Add label selector on cmd, analysis, server . so users can use `analyze` with label selectors

Closes #993  

## 📑 Description

grpc scheme should be merged and need to update scheme package.
https://github.com/k8sgpt-ai/schemas/pull/28

## ✅ Checks


cmd/analyze/analyze.go
```diff
@@ -142,4 +144,6 @@ func init() {
 	AnalyzeCmd.Flags().BoolVarP(&customAnalysis, "custom-analysis", "z", false, "Enable custom analyzers")
 	// add custom headers flag
 	AnalyzeCmd.Flags().StringSliceVarP(&customHeaders, "custom-headers", "r", []string{}, "Custom Headers, <key>:<value> (e.g CustomHeaderKey:CustomHeaderValue AnotherHeader:AnotherValue)")
+	// label selector flag
+	AnalyzeCmd.Flags().StringVarP(&labelSelector, "selector", "L", "", "Label selector (label query) to filter on, supports '=', '==', and '!='. (e.g. -l key1=value1,key2=value2). Matching objects must satisfy all of the specified label constraints.")
```
arbreezy, line 148: I think accepting a slice of label key pairs, will make the feature even more useful

pkg/analyzer/gateway.go
```diff
@@ -41,7 +43,15 @@ func (GatewayAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := client.List(a.Context, gtwList, &ctrl.ListOptions{}); err != nil {
+	labelSelectorMap := make(map[string]string)
```
arbreezy, line 46: Probably worth creating a util function since you replicate it for 3 analysers and is likely more CR analysers from the new gateway k8s api will be introduced  e.g TCPRoute

cmd/analyze/analyze.go
arbreezy: oh so you already support multiple key value pairs for labels !

awesome, then my only suggestion is to add a test case with more than one label just to sanity check it 👍

arbreezy: > What will it be like?
I was just  thinking that we can work with `StringSliceVarP` and have it as 
`./bin/k8sgpt analyze -L app.kubernetes.io/part-of=argocd app.kubernetes.io/name=argocd-application-controller` 

but if you already support multiple labels, I think is fine as it is.

pkg/analyzer/gateway.go
JuHyung-Son: good! I added on some analyzer

arbreezy (approved):
Great PR

AlexsJones (approved):

PR #1207, opened 2024-07-21
feat: add custom-analyzer cmd

matthisholleville:
Closes # 

## 📑 Description

Introduced the ability to manage custom analyzers in the K8sGPT application, enabling users to a
dd, deploy, and configure custom analyzers from various sources. This enhancement supports extending
 the application's analytical capabilities by integrating external analysis tools, thus offering mor
e flexibility and customization options to meet specific user needs.

## ✅ Checks

## ℹ Additional Information

https://github.com/user-attachments/assets/f7159272-4546-4030-a416-6dafd41546b1


arbreezy:
Great PR @matthisholleville, i think we should [support](https://docs.docker.com/engine/api/sdk/examples/#pull-an-image-with-authentication)private registries as well

matthisholleville:
@arbreezy Shouldn't we refer to daemon authentication ( docker login ) instead?

arbreezy:
> @arbreezy Shouldn't we refer to daemon authentication ( docker login ) instead?

Oh if it picks it up automatically, then it should be fine

matthisholleville:
Well spotted, @arbreezy ! Thanks to your comment, I noticed another issue. The version you reviewed never performed a `pull`. So, if the image was not available in the daemon's local registry, it could not be used. I updated it with a commit that pulls the image if it does not exist, while also adding authentication via username/password.

pkg/customAnalyzer/docker/docker.go
```diff
@@ -0,0 +1,136 @@
+package docker
```
AlexsJones, line 1: I am reluctant to take a hard dependency like docker :(  We need to discuss also if this is the approach we want to take

AlexsJones:
We probably need an example of at least one custom analyzer wrapped in a docker container before we merge this

AlexsJones:
@matthisholleville I love the PR and the enthusiasm. I'd like you to break it up into two pieces please.
One should be the commands just to add/remove/update/list the config file that has custom_analyzer in it.

In the second PR we can discuss the "--pull" flag mechanics (what you've done in docker).

e.g.

```
k8sgpt analyzer add security_scanner_linux -p 8085 -h localhost  # or k8sgpt analyzer add security_scanner_linux --pull --p 8085 -h localhost"

**Comment by matthisholleville** [comment]
@AlexsJones I removed Docker installation type

File: pkg/customAnalyzer/customAnalyzer.go [hunk 2]
```diff
@@ -0,0 +1,39 @@
+package customanalyzer
```
AlexsJones, line 1: Can we name this package and directory `custom_analyzer` ?

cmd/customAnalyzer/add.go
```diff
@@ -0,0 +1,73 @@
...
+	name string
+	url  string
+	port int
+)
+
+var addCmd = &cobra.Command{
+	Use:     "add",
+	Aliases: []string{"add"},
+	Short:   "This command will add a custom analyzer from source",
+	Long:    "This command allows you to add a custom analyzer from a specified source and optionally install it.",
```
AlexsJones, line 35:
```suggestion
This command allows you to add/remote/list an existing custom analyzer.
```

cmd/customAnalyzer/add.go
```diff
@@ -0,0 +1,73 @@
...
+			color.Red("Error writing config file: %s", err.Error())
+			os.Exit(1)
+		}
+		color.Green("%s added to the custom analyzers config list", name)
+
+	},
+}
+
+func init() {
+	addCmd.Flags().StringVarP(&name, "name", "n", "my-custom-analyzer", "Name of the custom analyzer.")
```
AlexsJones, line 70: A small point, but for these analyzers to show results in k8s they must validate against
`//regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')`

AlexsJones (changes requested):

pkg/customAnalyzer/customAnalyzer.go
matthisholleville: I'm not a big fan of snake case but make sense

AlexsJones (approved):

PR #1256, opened 2024-09-20
feat: error from events for STS analyzer

naveenthangaraj03:
Closes # 

## 📑 Description
- Need to get the error from event for statefulset analyzer, previously it doesn't get the clear error from this analyzer, in the below image there are two statefulset with 0/3 and 0/1 but it doesn't get the error.
Screenshot 2024-09-18 124356

- After getting the error from event for statefulset and if there is no event then it will check the pod is in running state or not
Screenshot 2024-09-19 165409

## ✅ Checks


pkg/analyzer/statefulset.go
```diff
@@ -93,6 +94,41 @@ func (StatefulSetAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
...
+			for i := int32(0); i < *(sts.Spec.Replicas); i++ {
+				podName := sts.Name + "-" + fmt.Sprint(i)
+				pod, err := a.Client.GetClient().CoreV1().Pods(sts.Namespace).Get(a.Context, podName, metav1.GetOptions{})
+				if err != nil {
+					if errors.IsNotFound(err) && i == 0 {
+						evt, err := util.FetchLatestEvent(a.Context, a.Client, sts.Namespace, sts.Name)
+						if err != nil || evt == nil || evt.Type == "Normal" {
+							break
+						}
+						failures = append(failures, common.Failure{
```
AlexsJones, line 107: Need someone else to take a look here

PR #1280, opened 2024-10-23
fix: add maxTokens to serve mode

samirtahir91:
## 📑 Description
We need to set `maxTokens` for serve mode, this is not set by default like it is with `k8sgpt auth`, which defaults to `2048`.
Backends like google gemini and googlevertexai fail without setting maxTokens when running in serve mode (so including k8sgpt operator).

This PR adds a new environment variable `K8SGPT_MAX_TOKENS` which you can set for serve, the default is  set to `2048` if unset

## ✅ Checks

## ℹ Additional Information
The k8sgpt operator will need to be updated to set maxTokens and providerId also in the k8sgpt CRD spec - injecting the env vars - see https://github.com/k8sgpt-ai/k8sgpt-operator/pull/545


samirtahir91:
@AlexsJones  Could you review please?

cmd/serve/serve.go
```diff
@@ -102,6 +103,22 @@ var ServeCmd = &cobra.Command{
...
+				env := os.Getenv("K8SGPT_MAX_TOKENS")
+				if env == "" {
+					return defaultMaxTokens
+				}
+				maxTokens, err := strconv.ParseInt(env, 10, 32)
+				if err != nil {
+					color.Red("Unable to convert maxTokens value: %v", err)
+					os.Exit(1)
+				}
+				if maxTokens < 10 || maxTokens > 32768 {
```
AlexsJones, line 116: Rather than setting the limit here, it probably works better for the specific checks inside the AI backend

AlexsJones (changes requested):

cmd/serve/serve.go
samirtahir91: Agree, happy if I remove the range check here?

Limits for max tokens per AI can be updated in a future PR, I think this should also apply to topk as it varies per backend (I followed the same generic check approach)

samirtahir91:
@AlexsJones - Are you happy to merge?

AlexsJones (approved):
Thanks for this, I think it's a good first step

samirtahir91:
> Thanks for this, I think it's a good first step

No worries 😃, once merged can you review the change for the operator to adopt this in https://github.com/k8sgpt-ai/k8sgpt-operator/pull/545

samirtahir91:
@AlexsJones Can you merge this please?

cmd/serve/serve.go
```diff
@@ -102,6 +103,18 @@ var ServeCmd = &cobra.Command{
 				}
 				return int32(topK)
 			}
+			maxTokens := func() int {
```
matthisholleville, line 106: Shouldn't we have a dedicated function? This function could be useful in CLI mode if the user wants to override the configuration with environment variables.

Moreover, I find that this improves readability.

samirtahir91: @AlexsJones can you comment

AlexsJones: I am okay with it like this, if we create an issue to refactor this file later on.

PR #1339, opened 2024-12-03
feat: enable to source customHeaders from the config file

muscionig:
## 📑 Description

This PR introduces the ability to read custom headers directly from the configuration file. The existing behavior is preserved: custom headers passed via the CLI are appended to those specified in the configuration file, which can be left empty.

Upon reviewing the source code, I noticed that `customHeaders` are sometimes "blanked" to prevent exposing sensitive data. https://github.com/k8sgpt-ai/k8sgpt/blob/c3f448693d7648a1a98dbf1f714464d1b0724f89/cmd/dump/dump.go#L60 This likely explains why the initial implementation did not support reading `customHeaders` from the configuration file. However, this limitation prevents the operator from leveraging `customHeaders` in configurations. 

This PR addresses the issue by enabling `customHeaders` to be defined in the configuration file, making it more flexible and useful for the operator. 

Please confirm if my understanding aligns with the intended behavior, and let me know if you would prefer I open an issue for tracking this change. Also, I noticed the documentation has a section regarding `customHeaders`, I'll be happy to update that one as well if this change is acceptable.

## ✅ Checks

## ℹ Additional Information

Here is an example of writing the headers in the configuration file:
```yaml
ai:
    providers:
        - name: openai
          ...
          customHeaders:
            - Header-1: value-header-1
            - Header-2: value-header-2
    defaultprovider: openai
kubeconfig: ""
kubecontext: ""
```

### DIFF
--- a/pkg/ai/iai.go
+++ b/pkg/ai/iai.go
@@ -119,7 +119,7 @@ type AIProvider struct {
 	TopK           int32         `mapstructure:"topk" yaml:"topk,omitempty"`
 	MaxTokens      int           `mapstructure:"maxtokens" yaml:"maxtokens,omitempty"`
 	OrganizationId string        `mapstructure:"organizationid" yaml:"organizationid,omitempty"`
-	CustomHeaders  []http.Header `mapstructure:"customHeaders"`
+	CustomHeaders  []http.Header `mapstructure:"customHeaders" yaml:"customHeaders,omitempty"`
 }
 
 func (p *AIProvider) GetBaseURL() string {
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -155,7 +155,7 @@ func NewAnalysis(
 
 	aiClient := ai.NewClient(aiProvider.Name)
 	customHeaders := util.NewHeaders(httpHeaders)
-	aiProvider.CustomHeaders = customHeaders
+	aiProvider.CustomHeaders = append(aiProvider.CustomHeaders, customHeaders...)
 	if err := aiClient.Configure(&aiProvider); err != nil {
 		return nil, err
 	}


AlexsJones:
Thanks for submitting this, it's not quite complete though.
- You need to modify the original Custom Resource with the new fields, regenerate it also. https://github.com/k8sgpt-ai/k8sgpt-operator/blob/main/api/v1alpha1/k8sgpt_types.go#L137
- Then you will need to modify what config we send to the deployment over the wire to k8sgpt deployment here https://github.com/k8sgpt-ai/schemas/blob/main/protobuf/schema/v1/server_config_service.proto
- This then gets merged and you pull those changes into k8sGPT and the operator independently. Then they both need to be able to set and get those new fields e.g., operator https://github.com/k8sgpt-ai/k8sgpt-operator/blob/main/pkg/client/config.go#L13


I know you'll probably want me to do this, so this serves half as a note to myself to come back to this PR and help

AlexsJones:
Coming back around on this one, sorry it sat so long. The two-line CLI change still reads fine (config-file headers preserved, CLI ones appended on top). What stalled it was my January ask about the operator CRD and schemas plumbing, and eighteen months of silence after that is on me as much as anyone.

Are you still interested in finishing it? If the cross-repo part is more than you want to take on, say so and we can look at landing the CLI-side behaviour on its own, with operator support tracked as a follow-up issue. If I don't hear back in a few weeks I'll assume you've moved on and tidy this up.

muscionig:
Hi @AlexsJones, I'd be happy to pick this up! I'll share half the blame for the silence.

I'll review the operator side and follow up soon!

AlexsJones:
Checking in on this, @muscionig, since it's been about six weeks since you said you'd look at the operator side.

No pressure either way. If the operator CRD work turns out bigger than expected, landing this two-line CLI change on its own is still an option, with operator support as a follow-up issue. Just say which way you'd like to go. If there's no reply by early October, we'll take the CLI-only route or close this.

PR #1365, opened 2025-01-23
fix: prevent npe by handling checking error in NewAnalysis call

pintohutch:
If there is an issue in creating the Analysis config when calling analysis.NewAnalysis, then we want to check before assigning the context to a potentially nil pointer.

Closes # 

## ✅ Checks


AlexsJones (approved):
Good spot thanks

PR #1366, opened 2025-01-31
fix: reverse hpa ScalingLimited error condition

antoinedeschenes:
Closes # 

## 📑 Description

k8sgpt currently outputs `- Error: the desired count is within the acceptable range` when an HPA is working properly. This is due to the "ScalingLimited" condition to be reversed:
```
      message: the desired count is within the acceptable range
      reason: DesiredWithinRange
      status: "False"
      type: ScalingLimited
```

This also fixes a typo in the analyzer's name not matching the `kind`, although that could be removed for another time. 

## ✅ Checks

### DIFF
--- a/pkg/analyzer/hpa.go
+++ b/pkg/analyzer/hpa.go
@@ -66,7 +66,6 @@ func (HpaAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 						Sensitive: []common.Sensitive{},
 					})
 				}
-				break
 			default:
 				if condition.Status == corev1.ConditionFalse {
 					failures = append(failures, common.Failure{


pkg/analyzer/analyzer.go
```diff
@@ -46,7 +46,7 @@ var coreAnalyzerMap = map[string]common.IAnalyzer{
 }
 
 var additionalAnalyzerMap = map[string]common.IAnalyzer{
-	"HorizontalPodAutoScaler": HpaAnalyzer{},
+	"HorizontalPodAutoscaler": HpaAnalyzer{},
```
antoinedeschenes, line 49: This would break `k8sgpt filters add`, maybe I should remove it?

AlexsJones: Are you saying because this is using v2 api it would break on default?

antoinedeschenes: I’m sorry, this wasn't clear. I mean that renaming the filter won’t align with the `active_filters` value used in an existing `k8sgpt.yaml` config file. This might need a config migration.

AlexsJones (approved):
Need to figure out if this needs us to update the config

pkg/analyzer/analyzer.go
AlexsJones: I'll merge this in - can you please validate after we push the new version? (v0.4.0)

antoinedeschenes: Sure thing

PR #1375, opened 2025-02-17
feat: adding flag filter for custom analyzer 

naveenthangaraj03:
Closes # 

## 📑 Description
ISSUES:
- If we run "k8sgpt analyze --custom-analysis" it will fetch all the all the custom analyzer as well as the active filters
- We need to run separate server for each custom analyzer
- Filter is not defined for custom analyzer

RESOLVED:
- If "--custom-analysis" is used it will get only the custom analyzer results, it will not include active filters
- If flag filter is specified for custom analyzer it will get the particular custom analyzer
- If flag filter is not specified in custom-analyzer it will get all the custom-analyzer results
- Passing filter name for custom analyzer server so that single server will handle multiple custom analyzer and run that particular custom analyzer resource based on switch case or if else

image

image

- In server we can write code logic based on the filter name

image

## ✅ Checks


AlexsJones:
@naveenthangaraj03 that's fine, go ahead and close this out and open with a single commit ( probably the easiest way )

pkg/custom/client.go
```diff
@@ -1,8 +1,7 @@
 package custom
 
 import (
-	rpc "buf.build/gen/go/k8sgpt-ai/k8sgpt/grpc/go/schema/v1/schemav1grpc"
-	schemav1 "buf.build/gen/go/k8sgpt-ai/k8sgpt/protocolbuffers/go/schema/v1"
+	rpc "github.com/naveenthangaraj03/k8sgpt-custom/proto/schema1/v1"
```
three-foxes-in-a-trenchcoat, line 4: This is the main blocker. We can't point k8sgpt's wire protocol at a personal fork. The reason for the swap is that the upstream `RunRequest` has no `Filter` field, so the right path is to add `filter` to the proto in `k8sgpt-ai/schema`, publish it, then bump the `buf.build/gen/go/...` dependency here and keep this import on the official package. Once that lands the rest of this change gets much cleaner.

go.mod
```diff
@@ -47,7 +47,10 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.25.1
 	github.com/hupe1980/go-huggingface v0.0.15
 	github.com/kyverno/policy-reporter-kyverno-plugin v1.6.4
+	github.com/naveenthangaraj03/k8sgpt-custom v0.0.0-20250217091002-d783b8c5750d
```
three-foxes-in-a-trenchcoat, line 50: Tied to the schema point above. This dependency should come out once the proto change is upstream.

go.mod
```diff
@@ -47,7 +47,10 @@ require (
...
+	github.com/metoro-io/mcp-golang v0.11.0
```
three-foxes-in-a-trenchcoat, line 51: This one looks unrelated to the feature and probably crept in by accident. Worth dropping to keep the PR focused.

go.mod
```diff
@@ -47,7 +47,10 @@ require (
...
 	github.com/mark3labs/mcp-go v0.36.0
+	github.com/naveenthangaraj03/k8sgpt-custom v0.0.0-20250217091002-d783b8c5750d
```
three-foxes-in-a-trenchcoat, line 53: Same module is already listed on line 50, so this is a duplicate require entry. `go mod tidy` should clear it up.

cmd/analyze/analyze.go
```diff
@@ -86,11 +86,13 @@ var AnalyzeCmd = &cobra.Command{
 			if verbose {
 				fmt.Println("Debug: All custom analyzers completed.")
 			}
+		} else {
```
three-foxes-in-a-trenchcoat, line 89: Worth calling out that this changes behavior. With the `else`, `--custom-analysis` now skips the core analyzers entirely, where before they ran alongside it. The issue does ask for this, so it may well be what we want, but since it is a behavior change it is worth a quick maintainer confirmation plus a docs note so existing users are not surprised.

pkg/analysis/analysis.go
```diff
@@ -247,48 +248,79 @@ func (a *Analysis) RunCustomAnalysis() {
...
-				fmt.Printf("Debug: %s launched.\n", cAnalyzer.Name)
-			}
-
-			result, err := canClient.Run()
-			if result.Kind == "" {
-				// for custom analyzer name, we must use a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.',
-				//and must start and end with an alphanumeric character (e.g. 'example.com',
-				//regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
-				result.Kind = cAnalyzer.Name
+	if len(a.Filters) != 0 {
```
three-foxes-in-a-trenchcoat, line 251: This filtered branch runs sequentially and appends to `a.Errors` and `a.Results` without taking the `mutex`, so it diverges from the concurrency model used in the unfiltered path just below. It would be cleaner to keep the single goroutine plus semaphore plus mutex loop and just add a name check inside it, rather than maintaining two copies of the logic. That also avoids the nested filter by analyzer loop.

pkg/analysis/analysis.go
```diff
@@ -247,48 +248,79 @@ func (a *Analysis) RunCustomAnalysis() {
...
+		for _, filter := range a.Filters {
+			for _, cAnalyzer := range customAnalyzers {
+				if cAnalyzer.Name == filter {
```
three-foxes-in-a-trenchcoat, line 254: If the filter does not match any custom analyzer name, nothing runs and nothing is reported, so the user just sees empty output. A short error or warning when a filter matches no analyzer would make this much easier to use, similar to the `filter does not exist` message the core path emits a bit further down.

pkg/custom_analyzer/customAnalyzer.go
```diff
@@ -34,12 +33,6 @@ func (*CustomAnalyzer) Check(actualConfig []CustomAnalyzerConfiguration, name, u
 			return fmt.Errorf("custom analyzer with the name '%s' already exists. Please use a different name", name)
```
three-foxes-in-a-trenchcoat, line 33: Removing the duplicate connection check changes existing validation behavior. If the intent is to let one server back several analyzers and dispatch by filter name, that makes sense, but it would help to keep that rationale in the PR description and ideally cover it with a test. As written there is also a now empty block left behind here.

pkg/analysis/analysis_test.go
```diff
@@ -623,15 +623,26 @@ func TestVerbose_RunCustomAnalysisWithCustomAnalyzer(t *testing.T) {
 
 	expectedOutputs := []string{
 		"Debug: Found custom analyzers [TestCustomAnalyzer].",
+	}
+
+	unexpectedOutputs := []string{
```
three-foxes-in-a-trenchcoat, line 628: This test now asserts the analyzer is not launched, but the unfiltered path should still launch it, so this reads as adjusting the test to match rather than verifying the new behavior. It would be great to instead add a test that exercises the filter path directly, asserting only the matching analyzer runs.

three-foxes-in-a-trenchcoat (commented):
Thanks for picking this up. The feature itself is one we want, and the overall direction of running only the selected custom analyzer is right. Left a few inline notes on the things that need sorting before this can merge. The biggest one is the schema dependency. Happy to help work through these.

PR #1479, opened 2025-04-30
feat: return Inference policy

ju187:
Closes # 

## 📑 Description
return the model id in the configure if no match

## ✅ Checks


.github/workflows/release.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - '[0-9]+.[0-9]+.x'
+      - inference-policy
```
AlexsJones, line 8: Can we remove this?

PR #1574, opened 2025-08-30
feat: add flag filter for custom analyzer

naveenthangaraj03:
Closes # 

## 📑 Description

ISSUES:

- Flag filter is not supported for custom-analyzer.
- If we have more than one custom analyzer need to run separate server for each custom analyzer.

RESOLVED:

- Flag filter is added for custom-analyzer, so we can define a lot of custom analyzer in a single server and run specific custom analyzer based on filter.

image

## ✅ Checks


naveenthangaraj03:
@AlexsJones , does this PR looks good?

naveenthangaraj03:
Hi @AlexsJones, I hope you're doing good. Could you please update me on this PR?

AlexsJones (changes requested):
Hey @naveenthangaraj03, thanks for working on this! Filter support for custom analyzers is a solid feature request and I can see the problem you're solving here. A few things to address before this can move forward:

**Personal protobuf dependency**

The switch from `buf.build/gen/go/k8sgpt-ai/k8sgpt` to `github.com/naveenthangaraj03/k8sgpt-custom` is a blocker. A core project shouldn't depend on a personal fork since it could diverge or disappear at any time. The `RunRequest` proto changes (adding the `filter` field) should be contributed upstream to the `k8sgpt-ai` protobuf schema first, then consumed here.

**Core analyzers get skipped when filters are set**

The `if/else` change in `cmd/analyze/analyze.go` means that passing `--filter` now runs *only* custom analyzers and skips core analyzers entirely. That's a breaking change for anyone currently using `--filter` to target built-in analyzers like `Pod` or `Service`. Filters should be partitioned between core and custom rather than making them mutually exclusive.

**Code duplication in RunCustomAnalysis**

The filtered and unfiltered paths are nearly identical (~40 lines each). You can avoid this by filtering the analyzer list upfront and then running the single existing concurrent loop over the filtered result:

```go
analyzersToRun := customAnalyzers
if len(a.Filters) != 0 {
    filterSet := make(map[string]bool)
    for _, f := range a.Filters {
        filterSet[f] = true
    }
    analyzersToRun = nil
    for _, ca := range customAnalyzers {
        if filterSet[ca.Name] {
            analyzersToRun = append(analyzersToRun, ca)
        }
    }
}
// then use the existing concurrent loop over analyzersToRun
```

This also fixes two secondary issues with the current filtered path: it runs synchronously (slower than it needs to be) and it appends to shared slices without mutex protection.

**Removed duplicate-connection validation**

Removing the URL+port uniqueness check in `customAnalyzer.go` makes sense if you want multiple analyzers behind one server, but it would be good to call that out explicitly in the PR description so reviewers know it's intentional.

**CLA**

The EasyCLA bot is flagging unsigned commits. You'll need to sign the CLA before this can be merged.

The core idea here is good and worth pursuing. Once the proto dependency is sorted out upstream and the filter logic is simplified, this will be in much better shape. Happy to help if you have questions on any of the above!

naveenthangaraj03:
Hi @AlexsJones, 

Do you have a rough timeline for when this protobuf change is expected to be merged? Knowing the ETA will help me coordinate my follow-up works. Thanks!

AlexsJones:
When the checks pass please let me know and we can proceed

PR #1603, opened 2026-01-20
feat: add openai responses backend and update ghcr workflows

Albe83:
## Summary
- add an OpenAI Responses backend with request/response handling and tests
- register the backend in defaults/docs for gpt-5+ models
- fix GHCR workflow auth/permissions and image name casing

## Testing
- Not run (not requested)

### DIFF
--- a/.github/workflows/build_container.yaml
+++ b/.github/workflows/build_container.yaml
@@ -17,7 +17,7 @@ on:
 env:
   GO_VERSION: "~1.23"
   IMAGE_NAME: "k8sgpt"
-  REGISTRY_IMAGE: ghcr.io/Albe83/k8sgpt
+  REGISTRY_IMAGE: ghcr.io/albe83/k8sgpt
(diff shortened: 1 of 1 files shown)


Albe83:
We introduced a separate backend (openairesponses) because the newest OpenAI models (reasoning models like gpt‑5/o1) require the Responses API and behave differently from the legacy chat/completions endpoint. Concretely:

the Responses API uses different parameters (e.g., max_output_tokens) and some legacy params like temperature/max_tokens can be rejected by these models;
the response shape is different (no choices[0].message), so we need dedicated parsing to reliably extract output text;
keeping the existing openai backend unchanged prevents regressions for users on legacy models or chat/completions‑compatible proxies.
By adding a new backend we can support modern models without breaking the current behavior, and offer a safe, gradual migration path: users on newer models can opt into openairesponses, everyone else stays untouched.

PR #1607, opened 2026-02-06
feat: add K8SGPT_PROVIDER_REGION env var support for serve command

putz612:
## 📑 Description

Add support for `K8SGPT_PROVIDER_REGION` environment variable in the serve command to allow configuring the provider region for backends like `googlevertexai` and `amazonbedrock`.

Currently, the serve command reads `K8SGPT_PROVIDER_ID` but not the region, which causes `googlevertexai` backend to fail with `RESOURCE_PROJECT_INVALID` error when used with the k8sgpt-operator because the region is not being passed to the AI provider configuration.

## Problem

When using k8sgpt with the operator and `googlevertexai` backend:
1. The operator sets the `region` field in the K8sGPT CR
2. The operator should pass this as `K8SGPT_PROVIDER_REGION` env var (separate PR needed for k8sgpt-operator)
3. However, `cmd/serve/serve.go` doesn't read this env var, so `ProviderRegion` is always empty
4. This causes `googlevertexai.go` to use the default region, which may not match the project configuration

## Solution

Read `K8SGPT_PROVIDER_REGION` from environment and set it in the `AIProvider` struct, similar to how `K8SGPT_PROVIDER_ID` is handled.

## Changes

- Added `providerRegion := os.Getenv("K8SGPT_PROVIDER_REGION")` to read the env var
- Added `ProviderRegion: providerRegion,` to the `AIProvider` struct initialization

## Related

- A corresponding PR will be needed in k8sgpt-ai/k8sgpt-operator to set `K8SGPT_PROVIDER_REGION` env var for `googlevertexai` backend

## ✅ Checks


putz612:
@AlexsJones I see that the operator change got merged in. Is there anything else I need to do to get this merged in?

AlexsJones:
Sign the CLA please

putz612:
Complete

three-foxes-in-a-trenchcoat (commented):
Small but useful addition. Provider region support via env var is consistent with other provider-specific config.

Note: Consider also supporting this in the auth add/update commands for consistency.

LGTM.

PR #1629, opened 2026-03-24
fix: amazonbedrockconverse claude models temp and topp

CradleKing24:
Closes # 

## 📑 Description
For latest Claude models on Amazon Bedrock only support Temperature OR Top P inputs. This is a stop-gap fix to support all Claude models.  A likely better fix would be to check if Temperature OR Top P is set explicitly and only pass those to the Converse API call.  If both are set by the user, pass back the AWS Error (existing functionality) as the error details that only one is allowed to be set.

## ✅ Checks

### DIFF
--- a/pkg/ai/amazonbedrockconverse.go
+++ b/pkg/ai/amazonbedrockconverse.go
@@ -61,6 +61,11 @@ func processError(err error, modelId string) error {
 	}
 }
 
+func isClaudeModel(modelId string) bool {
+	m := strings.ToLower(modelId)
+	return strings.Contains(m, "claude")
+}
+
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat (commented):
Good fix. The Claude model check correctly separates temperature-only from dual temp+topP inference configs.

Note: The if/else is redundant — both branches set temperature. Only the else branch sets topP. Consider:
```go
if !isClaudeModel(a.model) {
    infConfig.TopP = aws.Float32(a.topP)
}
```

LGTM.

CradleKing24:
@three-foxes-in-a-trenchcoat Resolved comment and consolidated redundant conditional

AlexsJones (approved):

PR #1635, opened 2026-04-14
feat: add support of gpt2giga

doppoluv:
## 📑 Description

Added support for the GPT2Giga proxy provider to integrate with the GigaChat API via OpenAI/Anthropic-compatible interfaces. Key changes include:
1. Created gpt2giga.go with the client implementation:
    - Acts as a proxy to forward requests to the GigaChat API using OpenAI format
    - Configure() method supports setting proxy endpoints and model parameters
    - GetCompletion() sends requests to the proxy endpoint and processes responses
2. Added the Gpt2GigaClient to the list of available providers in iai.go
3. Updated backend and passwordless provider lists

## ✅ Checks

### DIFF
--- a/pkg/ai/gpt2giga.go
+++ b/pkg/ai/gpt2giga.go
@@ -1,0 +1,110 @@
+package ai
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+const (
+	gpt2gigaClientName = "gpt2giga"
+	baseURL            = "http://localhost:8090"
+)
+
+type Gpt2GigaClient struct {
+	nopCloser
+
+	client      *http.Client
+	baseURL     string
+	model       string
+	temperature float32
+	topP        float32
+	// organizationId string
+}
+
+func (c *Gpt2GigaClient) Configure(config IAIConfig) error {
+	c.baseURL = config.GetBaseURL()
+	if c.baseURL == "" {
+		c.baseURL = baseURL
+	}
+
+	c.client = &http.Client{}
+	if proxyEndpoint := config.GetProxyEndpoint(); proxyEndpoint != "" {
+		proxyUrl, err := url.Parse(proxyEndpoint)
+		if err != nil {
+			return err
+		}
+		c.client.Transport = &http.Transport{
+			Proxy: http.ProxyURL(proxyUrl),
+		}
+	}
+
+	c.model = config.GetModel()
+	c.temperature = config.GetTemperature()
+	c.topP = config.GetTopP()
+
+	return nil
+}
+
+func (c *Gpt2GigaClient) GetCompletion(ctx context.Context, prompt string) (string, error) {
+	// Making a request in the OpenAI format
+	requestBody := map[string]interface{}{
+		"model": c.model,
+		"messages": []map[string]string{
+			{"role": "user", "content": prompt},
+		},
+		"temperature": c.temperature,
+		"top_p":       c.topP,
+	}
+
+	jsonBody, err := json.Marshal(requestBody)
+
+	if err != nil {
+		return "", fmt.Errorf("error marshaling request: %w", err)
+	}
+
+	// Sending a request
+	url := fmt.Sprintf("%s/chat/completions", c.baseURL)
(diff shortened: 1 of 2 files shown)


AlexsJones:
It would be cleaner to reuse the existing OpenAIClient with a custom base URL instead of adding a new gpt2giga provider, as it avoids duplicated logic and maintenance overhead. If a distinct client is required, consider fixing the variable shadowing, adding tests, and updating docs.

three-foxes-in-a-trenchcoat (commented):
New provider support. The OpenAI-compatible proxy pattern is well-implemented.

Notes:
1. Default baseURL is localhost:8090 — consider documenting this as a proxy endpoint that users need to set up themselves.
2. No error handling for non-200 response bodies (just returns status code). Consider including the response body in the error for debugging.
3. No support for system messages or conversation history (single-turn only). Document this limitation.

LGTM for a new provider.

AlexsJones:
Hi @doppoluv — steward check-in here. Alex left design feedback back in April that's still open:

> It would be cleaner to reuse the existing `OpenAIClient` with a custom base URL instead of adding a new gpt2giga provider… If a distinct client is required, consider fixing the variable shadowing, adding tests, and updating docs.

Since then the branch has only picked up merges from `main` — nothing addressing that feedback — so it's stalled. Two ways forward:

1. **Preferred (Alex's suggestion):** drop the dedicated `Gpt2GigaClient` and just point the existing `openai` backend at the proxy (`--baseurl http://localhost:8090`). gpt2giga already exposes the OpenAI `/chat/completions` shape your client targets, so this avoids a duplicated backend and its maintenance.
2. **Keep the dedicated client**, but then please: (a) rename the `url := fmt.Sprintf(...)` local in `GetCompletion` — it shadows the imported `net/url` package (a likely `golangci-lint` trigger, which is currently red); (b) add a unit test — the other backends in `pkg/ai` use an `httptest.Server` to cover the happy path plus the non-200 and empty-`choices` error branches — so `codecov/patch` goes green; and (c) add a short docs note for the new backend.

Separately, **EasyCLA is still unsigned**, which is a hard merge blocker regardless of the approach above — the sign link is in the failing check.

Happy to re-review as soon as you've picked a direction and CI is green. If we don't hear back, we'll likely close this as stale in a couple of weeks — just drop a note to keep it open.

AlexsJones (changes requested):
Thanks for this, and I am sorry it has been sitting since April with nothing but a label on it.

The shape is right. `Gpt2GigaClient` implements the same surface as the other providers, registering it in `clients`, `Backends` and `passwordlessProviders` is all three of the places a new backend has to appear, and passwordless is the correct call for a local proxy.

Things to fix before it can land.

**It conflicts with main.** Five months of `pkg/ai` have moved under it, so this needs a rebase before anything else is worth doing.

**`const baseURL` is too general a name for package scope.** `pkg/ai` already uses `baseURL` as a local variable in `ollama.go` (line 43) and `customrest.go` (line 52), and already has a package-level `defaultBaseURL` in `ollama.go`. A package const called `baseURL` shadows nothing today but reads as if it governs the whole package. Call it `gpt2gigaDefaultBaseURL`.

**No tests, and `codecov/patch` is red at 0% of the diff.** The two parts worth pinning are pure enough to test cheaply with `httptest.NewServer`: that `GetCompletion` posts the OpenAI-shaped body with model, temperature and top_p, and that it returns the first choice's content while turning an empty `choices` array and a non-200 into errors. That is the difference between "it worked on my machine" and something the project can keep working.

**golangci-lint is red**, though the run is from 2026-04-21 and the annotation only says "issues found", so please re-run it after the rebase rather than chase the old output.

One question rather than a request: the endpoint is built as `%s/chat/completions`, so with the default base URL the call goes to `http://localhost:8090/chat/completions`. Most OpenAI-compatible proxies expose `/v1/chat/completions`. Is the missing `/v1` right for gpt2giga, or is it expected to be carried in a configured base URL? Worth a comment in the code either way, since the next reader will wonder.

Two small ones: drop the commented-out `// organizationId string`, and since this adds a user-facing backend, the docs listing supported backends need the entry too.

The `http.Client{}` without a timeout is worth a thought, though I will not hold the PR on it: no client in `pkg/ai` sets one today, so you are consistent with the house style, and #1731 is open on exactly this class of hang.

PR #1648, opened 2026-05-12
feat: analyze previous logs for restarted containers

### LINKED ISSUE #1370: [Feature]: Enabling the Log Analyzer to fetch previous logs of a pod when it restarts
### Checklist

- [x] I've discussed this feature request in the [K8sGPT Slack](https://k8sgpt.slack.com) and got positive feedback

Yes

### Problem Description

In Scenarios when a pod restarts (like if you try going to the UI of the sock-shop micro services demo[public archived] and try to go to order , the front end pod restarts), k8sgpt log analyzer is not prompted to look into the previous logs of the pod and so no results are available - even though this error may recurr anytime the said action flow is repeated by a user . It will be nice if Log analyzer is prompted to look into the previous logs of a restarted pod (kubectl logs <podname> -n <namepace> --previous)

### Solution Description

When a pod restarts , log analyzer should be prompted to check the previous logs of the restarted pod and if analyzed positive generate a result

### Benefits

Enables to track errors that lead to restarting pods , which otherwise are missed 

### Potential Drawbacks

_No response_

### Additional Information

_No response_

officialasishkumar:
Closes #1370

## 📑 Description
This PR updates the Log analyzer to inspect previous container logs when Kubernetes reports that a container has restarted. Current logs are analyzed as before; previous logs are fetched only for containers with `RestartCount > 0`, and matching findings are labeled as previous container logs so the source is clear.

The optional previous-log fetch does not create a new analyzer failure if Kubernetes no longer has the previous log stream available. A regression test verifies that restarted containers request logs with `PodLogOptions.Previous = true`.

## ✅ Checks

## ℹ Additional Information
Validated with:
- `go test ./pkg/analyzer -run 'TestLogAnalyzer'`\n- `go test ./pkg/analyzer`\n- `go test ./...`\n- `go vet ./pkg/analyzer`

### DIFF
--- a/pkg/analyzer/log.go
+++ b/pkg/analyzer/log.go
@@ -49,37 +49,11 @@ func (LogAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 	// Iterate through each pod
 
 	for _, pod := range list.Items {
-		podName := pod.Name
 		for _, c := range pod.Spec.Containers {
 			var failures []common.Failure
-			podLogOptions := v1.PodLogOptions{
-				TailLines: &tailLines,
-				Container: c.Name,
-			}
-			podLogs, err := a.Client.Client.CoreV1().Pods(pod.Namespace).GetLogs(podName, &podLogOptions).DoRaw(a.Context)
-			if err != nil {
-				failures = append(failures, common.Failure{
-					Text: fmt.Sprintf("Error %s from Pod %s", err.Error(), pod.Name),
-					Sensitive: []common.Sensitive{
-						{
-							Unmasked: pod.Name,
-							Masked:   util.MaskString(pod.Name),
-						},
-					},
-				})
-			} else {
-				rawlogs := string(podLogs)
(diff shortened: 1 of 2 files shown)


three-foxes-in-a-trenchcoat (approved):
## Review

**Verdict: LGTM**

Nice feature. The refactoring is clean — extracting `analyzeLogs()` as a shared helper eliminates the duplicated log-fetching logic, and the new `containerHasRestarted()` guard ensures we only request previous logs when there's actually a reason to.

Key strengths:
- Previous log fetch is opt-in only (only when `RestartCount > 0`), so no performance hit for normal pods
- Graceful degradation: if previous logs aren't available anymore, it's silently skipped (not flagged as error)
- Clear labeling: findings from previous logs are prefixed with "previous container logs:" so the source is unambiguous
- Test verifies both current and previous log requests via mock action inspection

One small suggestion: the `reportLogErrors` boolean parameter could be named more descriptively (e.g., `failOnError`), but this is cosmetic.

PR #1670, opened 2026-06-30
fix: guard against empty conditions in Gateway/GatewayClass analyzers

### LINKED ISSUE #1669: bug: Gateway/GatewayClass analyzers panic on objects with no status conditions
## Expected Behavior

A Gateway or GatewayClass that has no `.status.conditions` yet (a freshly created object before its controller writes status, or a cluster with no Gateway controller installed) should be skipped by the analyzer with no failure reported, just like the PodDisruptionBudget analyzer already does after #664.

## Current Behavior

`k8sgpt analyze` crashes the entire run with:

```
panic: runtime error: index out of range [0] with length 0
```

The Gateway analyzer reads `gtw.Status.Conditions[0]` and the GatewayClass analyzer reads `gc.Status.Conditions[0]` with no length check, so any Gateway/GatewayClass that has not been reconciled yet brings down the whole analyze run.

- `pkg/analyzer/gateway.go:77`
- `pkg/analyzer/gatewayclass.go:56`

## Steps to Reproduce

1. Apply a `Gateway` (or `GatewayClass`) resource to a cluster that has no Gateway controller installed, so its `.status.conditions` stays empty.
2. Run `k8sgpt analyze`.
3. The command panics with `index out of range [0] with length 0` instead of reporting/skipping the resource.

This is also confirmed via the analyzer unit tests: reverting the length-check guard makes the empty-conditions test panic with `index out of range [0] with length 0`.

## Environment

- **k8sgpt version:** main (`go build` from current `main`)
- **Kubernetes version:** any (reproduces against the Gateway API types directly)
- **AI Backend/Provider:** N/A (panic happens during analysis, before any AI call)
- **OS/Platform:** any; reproduced on macOS with Go 1.24+

## Additional Context

The PodDisruptionBudget analyzer hit the same class of bug and was fixed in #664 by checking `len(...Status.Conditions)` before indexing `[0]`. The Gateway and GatewayClass analyzers never received that guard. Fix incoming that mirrors the PDB precedent.

anxkhn:
Closes #1669

## 📑 Description

The Gateway and GatewayClass analyzers index `Status.Conditions[0]` with no length check. A Gateway or GatewayClass with empty `status.conditions` (a brand-new object before its controller writes status, or a cluster with no Gateway controller installed) panics the whole `k8sgpt analyze` run with `index out of range [0] with length 0`.

This mirrors the fix the PodDisruptionBudget analyzer received in #664: short-circuit on `len(...Conditions) > 0` before reading `[0]`, so the object is simply skipped.

- `pkg/analyzer/gateway.go`: `if len(gtw.Status.Conditions) > 0 && gtw.Status.Conditions[0].Status != metav1.ConditionTrue`
- `pkg/analyzer/gatewayclass.go`: same guard for `gc.Status.Conditions[0]`

Added `TestEmptyConditionsGatewayAnalyzer` and `TestEmptyConditionsGatewayClassAnalyzer` (empty conditions -> 0 results, no panic). Both panic on the unpatched source, so they actually exercise the bug.

## ✅ Checks

## ℹ Additional Information

No behavior change for Gateways/GatewayClasses that already report conditions. `gofmt`, `go vet ./pkg/analyzer/...`, `go build`, and `go test ./pkg/analyzer/...` all pass.

### DIFF
--- a/pkg/analyzer/gateway.go
+++ b/pkg/analyzer/gateway.go
@@ -74,7 +74,7 @@ func (GatewayAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 
 		// Check only the current conditions
 		// TODO: maybe check other statuses Listeners, addresses?
-		if gtw.Status.Conditions[0].Status != metav1.ConditionTrue {
+		if len(gtw.Status.Conditions) > 0 && gtw.Status.Conditions[0].Status != metav1.ConditionTrue {
 			failures = append(failures, common.Failure{
 				Text: fmt.Sprintf("Gateway '%s/%s' is not accepted. Message: '%s'.",
 					gtwNamespace,
--- a/pkg/analyzer/gateway_test.go
+++ b/pkg/analyzer/gateway_test.go
@@ -91,6 +91,49 @@ func TestGatewayAnalyzer(t *testing.T) {
 
 }
 
+func TestEmptyConditionsGatewayAnalyzer(t *testing.T) {
+	ClassName := gtwapi.ObjectName("exists")
+	AcceptedStatus := metav1.ConditionTrue
+	GatewayClass := BuildGatewayClass(string(ClassName))
+
+	Gateway := BuildGateway(ClassName, AcceptedStatus, nil)
+	// A freshly created Gateway (or one whose controller has not reported yet)
+	// has no status conditions; the analyzer must not panic on it.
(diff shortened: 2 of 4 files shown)


AlexsJones (approved):
Approving: tested nil/empty-guard panic fix, follows existing analyzer idioms.

Issue #653, opened 2023-09-13
[Bug]: Panic when analyzing ingress that has no paths (w suggested fix)


testA113:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)

### K8sGPT Version

v0.3.14

### Kubernetes Version

v1.27

### Host OS and its Version

MacOS 14 Sonoma 

### Steps to reproduce

- Create a ingress that has no paths defined yet (this use case may arise when an expert wants to configure an ingress rule with annotations and TLS secrets, while leaving defining paths for services up to developers).
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: default-ingress-1
spec:
  ingressClassName: nginx
  rules:
  - host: asdf
 ```

- Run `k8sgpt analyze --filter Ingress`

### Expected behaviour

There should be a valid response with error message summaries or no errors.

### Actual behaviour

```sh
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10410719c]

goroutine 27 [running]:
github.com/k8sgpt-ai/k8sgpt/pkg/analyzer.IngressAnalyzer.Analyze({}, {0x14000703710, {0x104f0bc18, 0x14000126018}, {0x0, 0x0}, {0x104f0c350, 0x106207020}, 0x0, {0x0, ...}, ...})
	/home/runner/work/k8sgpt/k8sgpt/pkg/analyzer/ingress.go:102 +0xc3c
github.com/k8sgpt-ai/k8sgpt/pkg/analysis.(*Analysis).RunAnalysis.func2({0x104eefca0?, 0x10624c6b8?}, {0x14000054240, 0x7})
	/home/runner/work/k8sgpt/k8sgpt/pkg/analysis/analysis.go:198 +0xc0
created by github.com/k8sgpt-ai/k8sgpt/pkg/analysis.(*Analysis).RunAnalysis
	/home/runner/work/k8sgpt/k8sgpt/pkg/analysis/analysis.go:196 +0x7c0
```

### Additional Information

It could be worth checking if `rule.HTTP != nil` before looping through `role.HTTP.Paths` at `k8sgpt/pkg/analyzer/ingress.go:102`

PR #1672, opened 2026-06-30
fix: guard against nil TargetRef in Service not-ready endpoints

### LINKED ISSUE #1671: bug: Service analyzer panics on not-ready endpoint with nil TargetRef
## Expected Behavior

A Service whose Endpoints contain a not-ready address with no `TargetRef` should be reported as a single "Service has not ready endpoints" failure, like any other not-ready endpoint.

## Current Behavior

The Service analyzer dereferences `EndpointAddress.TargetRef` while building the not-ready pod list:

```go
// pkg/analyzer/service.go
for _, addresses := range epSubset.NotReadyAddresses {
    count++
    pods = append(pods, addresses.TargetRef.Kind+"/"+addresses.TargetRef.Name)
}
```

`TargetRef` is `*v1.ObjectReference` and is marked `+optional` in `k8s.io/api`; it is nil for endpoint addresses that are not backed by a Pod (bare-IP addresses or manually-created `Endpoints`). A single such not-ready address makes the whole `k8sgpt analyze` run panic instead of reporting the Service:

```
panic: runtime error: invalid memory address or nil pointer dereference
... pkg/analyzer/service.go:99
```

## Steps to Reproduce

1. Create a Service with a selector.
2. Create an `Endpoints` object for it whose `subsets[].notReadyAddresses` contains an address with only an `ip` set (no `targetRef`), e.g. a manually-managed or bare-IP endpoint.
3. Run `k8sgpt analyze`.
4. The run panics with a nil pointer dereference at `pkg/analyzer/service.go:99`.

## Environment

- **k8sgpt version:** main (`pkg/analyzer/service.go`)
- **Kubernetes version:** n/a (reproducible against the analyzer with `k8s.io/client-go/kubernetes/fake`)
- **AI Backend/Provider:** n/a (panic happens during analysis, before any backend call)
- **OS/Platform:** any

## Additional Context

Guarding `TargetRef != nil` before appending the pod label (while still counting the address) fixes the crash and keeps the failure message unchanged. This mirrors the existing nil/empty guards added in #653 and #664.

anxkhn:
Closes #1671

## 📑 Description

The Service analyzer dereferences `EndpointAddress.TargetRef` while building the not-ready pod list, but `TargetRef` is optional (`*v1.ObjectReference`, marked `+optional` in `k8s.io/api`) and is nil for endpoint addresses not backed by a Pod, for example bare-IP addresses or manually-created `Endpoints`. A single not-ready address of that kind panics the whole `k8sgpt analyze` run with `invalid memory address or nil pointer dereference` at `pkg/analyzer/service.go:99`.

This only appends the `Kind/Name` pod label when `TargetRef != nil`. The address is still counted, so the "Service has not ready endpoints" failure is reported exactly as before; the crash just goes away. This mirrors the existing nil/empty guards added in #653 and #664.

## ✅ Checks

## ℹ Additional Information

Adds a regression test (`service_test.go`) for a not-ready endpoint address with no `TargetRef`. Verified that reverting only the guard makes the new test panic at `pkg/analyzer/service.go:99`, and restoring it makes the suite pass. `go build`, `gofmt -l`, `go vet ./pkg/analyzer/...`, and `go test ./pkg/analyzer/...` all pass. No behavior change for Services that already had Pod-backed endpoints. Commit is DCO signed-off.

### DIFF
--- a/pkg/analyzer/service.go
+++ b/pkg/analyzer/service.go
@@ -96,7 +96,9 @@ func (ServiceAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 				if len(epSubset.NotReadyAddresses) > 0 {
 					for _, addresses := range epSubset.NotReadyAddresses {
 						count++
-						pods = append(pods, addresses.TargetRef.Kind+"/"+addresses.TargetRef.Name)
+						if addresses.TargetRef != nil {
+							pods = append(pods, addresses.TargetRef.Kind+"/"+addresses.TargetRef.Name)
+						}
 					}
 				}
 			}
--- a/pkg/analyzer/service_test.go
+++ b/pkg/analyzer/service_test.go
@@ -121,6 +121,53 @@ func TestServiceAnalyzer(t *testing.T) {
 			},
 		},
 		{
+			name: "Service with not ready endpoints with nil TargetRef",
+			config: common.Analyzer{
+				Client: &kubernetes.Client{
+					Client: fake.NewSimpleClientset(
+						&v1.Endpoints{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      "test-service",
+								Namespace: "default",
+							},
+							Subsets: []v1.EndpointSubset{
+								{
+									// Not-ready address with no TargetRef (bare-IP /
+									// manually-created Endpoints, not backed by a Pod).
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Approving: tested nil/empty-guard panic fix, follows existing analyzer idioms.

PR #1680, opened 2026-07-02
fix(ingress): only flag IngressClass as missing on NotFound errors

### LINKED ISSUE #1668: [Bug] Ingress analyzer false positive for alb IngressClass. PR #1599 fixed GKE classes but missed AWS ALB and other cloud-provider controllers
## Description

PR [#1599](https://github.com/k8sgpt-ai/k8sgpt/pull/1599) (fixing issue [#849](https://github.com/k8sgpt-ai/k8sgpt/issues/849)) introduced `isGKEBuiltInIngressClass()` to skip validation for GKE's `gce` and `gce-internal` classes. However, on AWS EKS with the AWS Load Balancer Controller, the IngressClass named `alb` is a real, present resource. The analyzer incorrectly reports it as missing and fires alerts on every analysis cycle.

## Steps to reproduce

1. Run an AWS EKS cluster with the AWS Load Balancer Controller installed. The `alb` IngressClass is created by the controller.
2. Create any Ingress resource using `ingressClassName: alb`.
3. Run `k8sgpt analyze` or observe operator Result CRs.
4. k8sgpt reports: _"Ingress uses the ingress class alb which does not exist."_

## Observed vs expected

**Actual:** k8sgpt reports _"Ingress uses the ingress class alb which does not exist"_ and creates Result CRs that fire alerts every 30 minutes.

**Expected:** k8sgpt does not flag the ingress. The `alb` IngressClass exists and the ingress is functioning correctly.

## Evidence — IngressClass exists

```
$ kubectl get ingressclass alb
NAME   CONTROLLER            PARAMETERS   AGE
alb    ingress.k8s.aws/alb          180d
```

The IngressClass resource exists in the cluster. This is the standard IngressClass created by the AWS Load Balancer Controller.

## Root cause — incomplete fix in PR #1599

The fix in #1599 hardcoded a GKE-specific allowlist. The `isGKEBuiltInIngressClass()` gate wraps the `IngressClasses().Get()` call, so when the allowlist does not include `alb`, the API call runs and (incorrectly) returns an error on some clusters.

## Suggested fix

Extend the existing allowlist introduced in PR #1599 to cover `alb`. 

```go
// rename for clarity and extend
func isImplicitIngressClass(className string) bool {
    switch className {
    case "gce", "gce-internal", // GKE — no IngressClass resource created by controller
        "alb":                   // AWS Load Balancer Controller
        return true
    }
    return false
}
```

## Related

- Issue #849 — original GKE report
- PR #1599 — partial fix (GKE only)

## Environment

- k8sgpt version: `v0.4.33`
- Deployment: k8sgpt-operator
- Kubernetes: AWS EKS
- Ingress controller: AWS Load Balancer Controller (`ingress.k8s.aws/alb`)
- IngressClass: `alb` (confirmed present via `kubectl get ingressclass alb`)

AlexsJones:
## Problem

The ingress analyzer (`pkg/analyzer/ingress.go`) treated **any** error from `IngressClasses().Get()` as proof the class was missing — including non-`NotFound` errors such as RBAC `Forbidden` or API timeouts. On clusters using cloud-provider controllers (e.g. the AWS Load Balancer Controller's `alb` class, per #1668) this reported a *present* IngressClass as missing and fired alerts every reconcile cycle.

Note: this is a more general fix than the allowlist approach suggested in the issue. Rather than extending the GKE-specific `isGKEBuiltInIngressClass` allowlist for every cloud provider (`alb`, `nginx`, `traefik`, …), we simply stop misreporting non-`NotFound` errors — which covers all controllers at once. The GKE allowlist is retained, since `gce`/`gce-internal` create no IngressClass resource at all.

## Fix

Gate the failure on `apierrors.IsNotFound(err)` so only a genuinely absent IngressClass is reported.

## Tests

- Added `TestIngressAnalyzerIngressClassGetError`:
  - a present `alb` IngressClass reports no error;
  - a `Forbidden` error (via a fake reactor) is **not** misreported as a missing IngressClass.
- Existing GKE/non-existent-class tests unchanged and passing.

Fixes #1668

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/analyzer/ingress.go
+++ b/pkg/analyzer/ingress.go
@@ -19,6 +19,7 @@ import (
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/util"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
@@ -83,7 +84,12 @@ func (IngressAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 			// an IngressClass resource (they are recognized by the GKE ingress controller)
 			if !isGKEBuiltInIngressClass(*ingressClassName) {
 				_, err := a.Client.GetClient().NetworkingV1().IngressClasses().Get(a.Context, *ingressClassName, metav1.GetOptions{})
-				if err != nil {
+				// Only flag the ingress when the IngressClass genuinely does not
+				// exist. Any other error (e.g. RBAC Forbidden, API timeouts) must
+				// not be misreported as a missing IngressClass, which produced
+				// false positives for cloud-provider classes such as the AWS Load
+				// Balancer Controller's "alb". See issue #1668.
+				if apierrors.IsNotFound(err) {
 					doc := apiDoc.GetApiDocV2("spec.ingressClassName")
 
 					failures = append(failures, common.Failure{
--- a/pkg/analyzer/ingress_test.go
+++ b/pkg/analyzer/ingress_test.go
@@ -15,6 +15,7 @@ package analyzer
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


AlexsJones:
Closing as a duplicate of #1713, which is the IngressClass IsNotFound fix (DCO/CI clean).

PR #1683, opened 2026-07-06
fix: guard against nil spec.replicas in deployment analyzer

anxkhn:
## Description
The Deployment analyzer dereferences `*deployment.Spec.Replicas` without a nil
check. `Spec.Replicas` is a `*int32`; although the API server usually defaults
it to 1, a Deployment object whose `spec.replicas` is explicitly unset (nil)
panics the analyze run with a nil pointer dereference at
`pkg/analyzer/deployment.go:57`.

This guards the comparison with a nil check, mirroring the sibling StatefulSet
analyzer (`pkg/analyzer/statefulset.go:97`), which already checks
`Spec.Replicas != nil` before dereferencing. The single guard also protects the
subsequent `*deployment.Spec.Replicas` reads on lines 58, 62, and 79, since they
only run inside the guarded block.

Added a regression test, `TestDeploymentAnalyzerNilReplicas`, that analyzes a
Deployment with nil `spec.replicas` and asserts `Analyze` returns no results and
does not panic.

## Checks

## Additional Information
No behavior change for Deployments that set `spec.replicas` (the overwhelming
common case, since the API server defaults it). The only difference is that a
Deployment with an explicitly-unset replicas field is now skipped instead of
crashing the analyzer.


AlexsJones (approved):
Reviewed the diff — this is a correct, minimal fix. ✅

- `pkg/analyzer/deployment.go:57` did dereference `*deployment.Spec.Replicas` unconditionally, and `Spec.Replicas` is a `*int32`, so a Deployment with an explicitly-unset `spec.replicas` would panic the analyze run. Real bug.
- The guard `deployment.Spec.Replicas != nil && …` short-circuits before the deref, and every other `*deployment.Spec.Replicas` read (the `Status.Replicas > *Spec.Replicas` comparison and both failure messages) lives inside that same block, so the single guard covers all of them.
- It mirrors the sibling `StatefulSetAnalyzer`, which already uses `sts.Spec.Replicas != nil && *(sts.Spec.Replicas) != …` — so this keeps the two analyzers consistent.
- `TestDeploymentAnalyzerNilReplicas` is a proper regression test: it feeds a nil-replicas Deployment and asserts no panic / zero results. CI (build, golangci-lint, validate, codecov) is green.

No behavior change for the common case where `spec.replicas` is set. Recommending this for merge — over to @AlexsJones for the final call.

---

PR #1685, opened 2026-07-07
test: assert hpa resource-configured error after the results loop

anxkhn:
Closes # 

## 📑 Description

`TestHPAAnalyzerWithExistingScaleTargetRefWithoutSpecifyingResources` in
`pkg/analyzer/hpa_test.go` cannot fail on the very regression it is meant to
guard against.

The test builds an HPA whose `ScaleTargetRef` points at an existing Deployment
whose only container declares no resource requests or limits, and expects the
HPA analyzer to report `... does not have resource configured.`. It verifies
that with:

```go
var errorFound bool
for _, analysis := range analysisResults {
    for _, err := range analysis.Error {
        if strings.Contains(err.Text, "does not have resource configured.") {
            errorFound = true
            break
        }
        if errorFound {   // dead: always false when reached
            break
        }
    }
    if !errorFound {      // nested INSIDE the outer loop
        t.Error("expected error 'does not have resource configured.' not found in analysis results")
    }
}
```

Because the `if !errorFound { t.Error(...) }` assertion sits inside the
`for _, analysis := range analysisResults` loop, it only runs when there is at
least one result. If the detection ever regresses and the analyzer returns zero
results, `analysisResults` is empty, the outer loop body never executes, the
assertion is skipped, and the test passes green, missing the exact failure it
exists to catch. The `if errorFound { break }` inside the inner loop is also
dead code (it can only be reached on an iteration that did not set `errorFound`).

The three sibling HPA tests already handle this correctly by placing the
assertion after the outer loop:
`TestHPAAnalyzerWithUnsuportedScaleTargetRef`,
`TestHPAAnalyzerWithNonExistentScaleTargetRef`, and the min-replicas test.

This change moves the assertion out of the outer loop, breaks out of the outer
loop once the error is found, and drops the dead inner-loop break, so the test
now matches its siblings and fails when the analyzer stops reporting the error.
It is a test-only change; no analyzer/production code is touched.

Verified: with the fix, breaking the analyzer's resource check so it emits zero
results turns this test red as intended, whereas the previous structure stayed
green on the same break. The full `pkg/analyzer` suite passes, and `gofmt`,
`go vet`, and `go build` are clean.

## ✅ Checks

## ℹ Additional Information

Test-only change, no behavior or public API change. Diff is 5 insertions /
5 deletions in a single file (`pkg/analyzer/hpa_test.go`).


AlexsJones (approved):
Reviewed the diff — correct fix to a false-negative in the test's own logic. ✅

- `TestHPAAnalyzerWithExistingScaleTargetRefWithoutSpecifyingResources` previously ran `if !errorFound { t.Error(...) }` **inside** the outer results loop, so it failed on the first analysis result that didn't contain the `does not have resource configured.` message — even when a later result did. The misplaced inner `if errorFound { break }` compounded it.
- The change hoists the final assertion out of both loops (assert the expected error was found in *any* result) and moves the early-out to break the outer loop once found. That's the correct semantic for a "some result contains X" check.
- Pure test refactor, +5/-5, no production code touched; CI green.

---

PR #1686, opened 2026-07-07
fix: resolve EKS default kubeconfig path across platforms

anxkhn:
Closes # 

## 📑 Description

The EKS integration analyzer builds the default kubeconfig path from
`os.Getenv("HOME")` when no explicit `--kubeconfig` is provided:

```go
kubeconfig = filepath.Join(os.Getenv("HOME"), ".kube", "config")
```

On Windows the `HOME` environment variable is normally unset (Windows exposes
`USERPROFILE`, and `HOMEDRIVE` plus `HOMEPATH`), so `os.Getenv("HOME")` returns
`""` and `filepath.Join("", ".kube", "config")` collapses to the relative path
`.kube\config`, resolved against the process's current working directory instead
of the user's home directory. `clientcmd` then loads an empty/non-existent
config, the current context is empty, and the analyzer reports `EKS cluster was
not detected` even when a valid `~/.kube/config` exists. In practice EKS
detection never works on Windows unless `--kubeconfig` is passed explicitly.

This changes the fallback to resolve the home directory with `os.UserHomeDir()`,
which is the cross-platform convention already used elsewhere in the codebase
(for example `cmd/root.go`). An explicit `--kubeconfig` still takes precedence.
The path resolution is extracted into a small `getKubeconfigPath` helper so it
can be unit tested, and the returned error is propagated through `Analyze`
(which already returns `([]common.Result, error)`).

## ✅ Checks

## ℹ Additional Information

No behavior change on Unix, where `os.UserHomeDir()` returns `$HOME`. The new
`getKubeconfigPath` is covered by `TestGetKubeconfigPath` in
`pkg/integration/aws/eks_test.go` (explicit `--kubeconfig` wins; otherwise the
fallback resolves to an absolute `<home>/.kube/config` rather than a relative
path). This is the only `os.Getenv("HOME")` in the repository.


AlexsJones (approved):
Reviewed the diff — correct cross-platform fix with a proper test. ✅

- `EKSAnalyzer.Analyze` built the default kubeconfig path from `os.Getenv("HOME")`. On Windows `HOME` is typically unset (Windows exposes `USERPROFILE`, and `HOMEDRIVE`+`HOMEPATH`), so `filepath.Join("", ".kube", "config")` collapsed to a relative `.kube\config` resolved against CWD; `clientcmd` then loaded an empty config and the analyzer always reported "EKS cluster was not detected" unless `--kubeconfig` was passed. Real bug.
- The fix routes through a new `getKubeconfigPath()` helper that resolves the home dir via `os.UserHomeDir()` — the cross-platform convention already used in `cmd/root.go` — while an explicit `--kubeconfig` (viper) still takes precedence. The error is propagated through `Analyze`, which already returns `([]common.Result, error)`.
- `TestGetKubeconfigPath` covers both branches (explicit path wins; otherwise the fallback resolves to an absolute `<home>/.kube/config`, asserting `filepath.IsAbs`), guarding the exact Windows regression. No behavior change on Unix, where `os.UserHomeDir()` returns `$HOME`.
- CI green (golangci-lint, build, validate, codecov).

---

PR #1687, opened 2026-07-07
test: assert Serve error synchronously after shutdown in TestServe

anxkhn:
`TestServe` (`pkg/server/server_test.go`) never actually verified that `Serve()`
returned cleanly on shutdown. The server was started in a goroutine that captured
`Serve()`'s return value, slept two seconds, then called `assert.NoError` on it:

```go
go func() {
    err := s.Serve()
    time.Sleep(time.Second * 2)
    assert.NoError(t, err, "Serve should not return an error")
}()
```

The test body returns immediately after `s.Shutdown()` without ever joining that
goroutine (no `WaitGroup`, channel, or other synchronization). So the assertion
runs roughly two seconds *after* `TestServe` has already been recorded as passed,
and can never affect the result. A live run finishes in about `0.00s`
(`go test ./pkg/server/ -run TestServe -v` -> `--- PASS: TestServe (0.00s)`),
i.e. well before the guarded assertion could execute.

The assertion was also asserting the wrong thing. `Shutdown()`
(`pkg/server/server.go:77-79`) closes the listener directly
(`s.listener.Close()`), and for the `EnableHttp: false` path `Serve()` runs
`grpcServer.Serve(lis)` and only ignores `http.ErrServerClosed`
(`pkg/server/server.go:138-142`). Closing the listener therefore makes `Serve()`
return a non-nil error that wraps `net.ErrClosed`, so the original
`assert.NoError` expected the opposite of the real behavior.

This change captures `Serve()`'s error over a buffered channel and asserts it
synchronously in the test body after `Shutdown()`, checking that it wraps
`net.ErrClosed` (guarded by a 5s timeout). Now the check runs within the test's
lifetime and actually catches an unexpected `Serve()` failure. It is a test-only
change; no production code is touched, and no new imports are needed (`net`,
`time`, `testify/assert` are already imported).

## Checks

## Additional Information

- Scope: one file, `pkg/server/server_test.go` (+12/-3). No behavior change to
  the server itself; this only makes an existing test assertion effective.
- Verification:
  - `go test ./pkg/server/ -run TestServe -count=1 -v` -> PASS.
  - `go test ./pkg/server/... -count=1` -> ok. `-race` on `TestServe` -> ok
    (the channel also removes the previous goroutine/test-body data race on the
    `err` local).
  - `gofmt -l pkg/server/server_test.go` -> empty; `go vet ./pkg/server/...` and
    `go build ./...` clean.
  - Red/green confirmation: temporarily pre-binding the port so `Serve()`'s
    `net.Listen` fails with `address already in use` (not `net.ErrClosed`) makes
    the new synchronous `assert.ErrorIs` fail, proving the check now reacts to an
    unexpected `Serve()` error within the test's lifetime. The old sleep + assert
    form did not. (Probe removed; not part of the diff.)


AlexsJones (approved):
Reviewed — this is a real flaky-test fix, not just a style change. The old goroutine did `Serve(); sleep(2s); assert.NoError(err)`: (a) a testify assertion in a goroutine can fire after the test has completed (panics on a dead *testing.T), and (b) `assert.NoError` is the wrong contract — `Shutdown()` is `listener.Close()` (pkg/server/server.go:78), so `Serve` legitimately returns a non-nil accept error wrapping `net.ErrClosed` after shutdown. The new synchronous `select` on a buffered channel with `assert.ErrorIs(err, net.ErrClosed)` plus a 5s timeout asserts exactly the right thing. `net` was already imported; test-only +12/-3; CI fully green (lint, build, codecov). Recommending merge.

PR #1691, opened 2026-07-08
fix: resolve shared client data race in concurrent grpc handler

glatinone:
While analyzing k8sgpt's gRPC server mode (`k8sgpt serve`), we identified a critical data race in the client creation path where concurrent `Query` requests can mutate shared client configurations.

**The Issue:**
- The factory currently returns shared pointer singletons from a package-level global slice.
- When concurrent gRPC requests invoke `factory.NewClient(backend)` followed by `.Configure(&config)`, they modify the shared struct's internal client, model, and endpoint configurations concurrently without any synchronization.
- This leads to silent configuration contamination where one user's query parameters (or API tokens) can bleed into another active request's session under load.

**The Fix:**
- Replaced the global shared client slice with a lazy per-call constructor registry map.
- `NewClient` now returns a fresh client instance pointer for every invocation, guaranteeing complete request isolation within the concurrent gRPC server path.

### DIFF
--- a/pkg/ai/concurrency_test.go
+++ b/pkg/ai/concurrency_test.go
@@ -1,0 +1,183 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ai
+
+import (
+	"math/rand"
+	"net/http"
+	"reflect"
+	"sync"
+	"testing"
+)
+
+type testCfg struct {
+	password        string
+	model           string
+	baseURL         string
+	proxyEndpoint   string
+	endpointName    string
+	engine          string
+	temperature     float32
+	providerRegion  string
+	topP            float32
+	topK            int32
+	maxTokens       int
+	stopSequences   []string
+	providerID      string
+	compartmentID   string
+	organizationID  string
+	azureAPIType    string
+	azureAPIVersion string
+	customHeaders   []http.Header
+}
+
+func (c *testCfg) GetPassword() string             { return c.password }
+func (c *testCfg) GetModel() string                { return c.model }
+func (c *testCfg) GetBaseURL() string               { return c.baseURL }
+func (c *testCfg) GetProxyEndpoint() string         { return c.proxyEndpoint }
+func (c *testCfg) GetEndpointName() string          { return c.endpointName }
+func (c *testCfg) GetEngine() string                { return c.engine }
+func (c *testCfg) GetTemperature() float32          { return c.temperature }
+func (c *testCfg) GetProviderRegion() string        { return c.providerRegion }
+func (c *testCfg) GetTopP() float32                 { return c.topP }
+func (c *testCfg) GetTopK() int32                   { return c.topK }
+func (c *testCfg) GetMaxTokens() int                { return c.maxTokens }
+func (c *testCfg) GetStopSequences() []string       { return c.stopSequences }
+func (c *testCfg) GetProviderId() string            { return c.providerID }
+func (c *testCfg) GetCompartmentId() string         { return c.compartmentID }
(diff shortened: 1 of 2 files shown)


AlexsJones (changes requested):
Thanks @glatinone — this is a real bug and the fix is the right shape. 👍

You've correctly spotted that `NewClient` handed every caller a pointer into the package-level `clients` slice of shared singletons, so under `k8sgpt serve` two concurrent gRPC `Query` requests would each call `Configure()` on the **same** instance and stomp one another's model/endpoint/credentials. Switching to a `clientConstructors map[string]func() IAI` that returns a fresh instance per call closes that race cleanly, and doing it by per-call isolation rather than a lock is the better choice here — no lock, no deadlock surface. I checked that each backend's `GetName()` still equals its new map key, so provider resolution is unchanged.

Two things before this can go in:

1. **Tests.** This is a concurrency fix landing with no test coverage. Please add (a) a simple isolation test — `NewClient("openai")` called twice returns distinct pointers — and (b) a concurrency test that runs `NewClient` + `Configure` from several goroutines with distinct configs and asserts no cross-contamination, run under `-race` so a future refactor can't silently reintroduce the shared singleton.

2. **CLA.** EasyCLA is currently failing — it'll block merge regardless of the code. Please sign it.

Two optional follow-ups while you're in here:
- `Backends` is still a hand-maintained slice parallel to `clientConstructors`; deriving or validating one against the other would stop them drifting.
- The `default` branch returns `&OpenAIClient{}` for an unknown provider name, which silently masks a misconfigured backend. Returning an error (or a no-op client) would surface the misconfig — pre-existing, but a natural time to fix it.

AlexsJones (approved):
Delta re-review of the new commit — both asks from the previous review are delivered. ✅

- `TestNewClient_ReturnsDistinctPointers` covers the isolation contract directly.
- `TestConcurrent_NewClient_Configure_NoBleed` hammers `NewClient` + `Configure` from 200 goroutines with distinct configs and asserts no two workers ever share a client pointer — under `-race` this permanently guards against a future refactor reintroducing the shared singleton.
- EasyCLA is now signed and the full CI matrix is green.

The two optional follow-ups (deriving `Backends` from `clientConstructors` so they can't drift, and surfacing an error instead of silently defaulting unknown providers to `OpenAIClient`) remain open invitations — fine as a separate PR.

Approving — recommending to the maintainers for final look and merge. Thanks for the quick turnaround, @glatinone.

AlexsJones:
Good news @glatinone — Alex has decided to merge this PR (the overlapping #1704 was closed as superseded in its favor). One thing stands between it and the merge button: the branch is behind `main` and this repo's protection requires branches to be up to date, and since the fork doesn't have "allow edits by maintainers" enabled we can't update it from this side. Could you rebase (or merge `main` into) `feature/client-concurrency-race-fix` and push? Once CI is green at the new head it merges straight away.

glatinone:
Quick update: I attempted to run the GitHub “Update branch” API, but the org’s OAuth App restrictions blocked it. I’ll rebase feature/client-concurrency-race-fix onto the latest main locally and push from my side. If you prefer, enabling this app or clicking “Update branch” on your side would also clear the gate. Thank you!

glatinone:
Maintainers: unable to use update-branch via API due to org OAuth app restrictions (403). Please click “Update branch” or temporarily allow this app so we can fast-forward the PR branch to main and clear the mergeable_state=dirty. CI is green otherwise. Thanks!

AlexsJones:
Thanks for trying the update-branch route. The situation has moved past what that button can do though: the branch now has real merge conflicts with `main` (mergeable state is dirty, not just behind), and Update branch only works for clean merges, so no permission change on our side would help. A local rebase of `feature/client-concurrency-race-fix` onto `main` and a force-push is the one path that clears it. CI is green otherwise and the approval stands, so this merges as soon as the rebased head is up.

AlexsJones:
Thanks for pushing the rebase through! I checked the rebased head against the approved change and it is exactly the same fix (constructor map in `pkg/ai/iai.go` plus the two isolation tests), with the conflicts against main now resolved. DCO, EasyCLA, License and validate are all green at 532d1bd, so this is finally conflict-free and ready for the merge.

glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

AlexsJones:
Thanks again for pushing the rebase through, and sorry this one sat so long, that's on me, not you. Closing wasn't necessary: the PR was approved and next in my merge queue. I checked the rebased head yesterday morning and it matched the approved change exactly, all checks green. If you reopen it I'll merge it as is. If you'd rather leave it closed, I'll re-land the fix with credit to you, but reopening is the easy path.

PR #1699, opened 2026-07-13
fix: guard empty API key in Google GenAI client

anxkhn:
Closes #

## 📑 Description
`GoogleGenAIClient.Configure` reads the first byte of the password token to detect
a JSON service-account credential:

```go
token := config.GetPassword()
authOption := option.WithAPIKey(token)
if token[0] == '{' {
    authOption = option.WithCredentialsJSON([]byte(token))
}
```

The `google` backend is not in `passwordlessProviders`, so a password is expected,
but nothing enforces a non-empty value. When the stored/entered API key is empty,
`token[0]` indexes an empty string and `Configure` panics:

```
panic: runtime error: index out of range [0] with length 0
```

This adds a `len(token) > 0` guard before indexing, matching how the sibling
clients handle an empty key (`pkg/ai/watsonxai.go` checks `apiKey == ""` and
`pkg/ai/anthropic.go` checks `token != ""` before using it). With an empty
password, `Configure` now falls through to `option.WithAPIKey("")` and returns the
Google SDK's normal authentication error instead of crashing.

A regression test (`pkg/ai/googlegenai_test.go`) asserts that `Configure` with an
empty password does not panic and returns an error. It fails with the exact
`index out of range [0] with length 0` panic without the guard and passes with it.

## ✅ Checks

## ℹ Additional Information
No behavior change for a valid API key or JSON credential; the guard only affects
the previously-panicking empty-key path. Verified locally with
`go test ./pkg/ai/...` (pass), `go vet ./pkg/ai/...`, `gofmt -l`, and `go build .`.

### DIFF
--- a/pkg/ai/googlegenai.go
+++ b/pkg/ai/googlegenai.go
@@ -41,7 +41,7 @@ func (c *GoogleGenAIClient) Configure(config IAIConfig) error {
 	// Access your API key as an environment variable (see "Set up your API key" above)
 	token := config.GetPassword()
 	authOption := option.WithAPIKey(token)
-	if token[0] == '{' {
+	if len(token) > 0 && token[0] == '{' {
 		authOption = option.WithCredentialsJSON([]byte(token))
 	}
 
--- a/pkg/ai/googlegenai_test.go
+++ b/pkg/ai/googlegenai_test.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Verified the panic path: `Configure` indexes `token[0]` on the raw password and the `google` backend isn't in `passwordlessProviders`, but nothing upstream enforces a non-empty value — so an empty stored key panics exactly as described. The `len(token) > 0` guard matches how the watsonx and anthropic clients already handle this, and the regression test locks in error-not-panic behavior.

Code is ready, but two checks block the merge:

- **EasyCLA fails on this PR only** (your other six from last night all pass) — "Missing CLA Authorization" usually means one commit here was authored with an email that isn't covered by your CLA signature. Could you authorize it via the EasyCLA link on the check, or rebase with the covered email?
- codecov/patch is coverage noise on a 1-line guard; not a concern.

Approving the code — once EasyCLA is green this is ready for @AlexsJones's final look.

anxkhn:
easycla and dco are green now. i also rebased onto main and added focused coverage for empty, api key, and json credential paths. @alexsjones, could you take another look when convenient?

AlexsJones (approved):
Thanks for the quick turnaround on the rebase, and the focused tests are exactly the coverage this needed (empty, API key, and JSON credential paths). Still LGTM at the new head.

anxkhn:

PR #1700, opened 2026-07-13
fix: clamp max-concurrency in custom analysis to avoid deadlock and panic

anxkhn:
Closes # 

`RunCustomAnalysis` built its worker semaphore straight from the unvalidated
`a.MaxConcurrency` field:

```go
semaphore := make(chan struct{}, a.MaxConcurrency)
```

`RunAnalysis` does not do this: it clamps the value first. Because
`--max-concurrency` is a user-supplied flag (`cmd/analyze`, default 10, no
lower-bound validation), a non-positive value passed into custom analysis caused
two failures:

- `k8sgpt analyze --custom-analysis --max-concurrency 0` builds an unbuffered
  channel. The dispatch path does `semaphore <- struct{}{}` and only launches the
  draining goroutine on the next line, so the send blocks on a receiver that is
  never reached: the whole `analyze` command deadlocks whenever a custom analyzer
  is configured.
- `k8sgpt analyze --custom-analysis --max-concurrency -1` makes
  `make(chan struct{}, -1)` panic with `makechan: size out of range`, crashing the
  process (this path is reached even with zero custom analyzers, since the `make`
  precedes dispatch).

The fix applies the same lower/upper-bound clamp that `RunAnalysis` already uses
(default 10 for non-positive values, cap at 100) before creating the semaphore.
There is no behavior change for valid positive values in 1..100.

Regression tests cover both cases:
- `TestRunCustomAnalysisZeroConcurrency` runs `RunCustomAnalysis` in a goroutine
  guarded by a timeout and fails on deadlock.
- `TestRunCustomAnalysisNegativeConcurrency` asserts it does not panic.

## Checks

## Additional Information

The added clamp is a verbatim mirror of the block already present in `RunAnalysis`
(`pkg/analysis/analysis.go`), including the inline comments and the
`const maxAllowedConcurrency = 100`, so the two paths now behave identically for
out-of-range concurrency. No public API or flag default changes.

---

## Files changed

```
 pkg/analysis/analysis.go      | 11 ++++++++++-
 pkg/analysis/analysis_test.go | 43 +++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 53 insertions(+), 1 deletion(-)
```

- `pkg/analysis/analysis.go`: adds the 8-line concurrency clamp before the
  `semaphore := make(...)` in `RunCustomAnalysis` (mirrors `RunAnalysis`).
- `pkg/analysis/analysis_test.go`: adds `time` import + two regression tests.

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -260,7 +260,16 @@ func (a *Analysis) RunCustomAnalysis() {
 		return
 	}
 
-	semaphore := make(chan struct{}, a.MaxConcurrency)
+	// Set a reasonable maximum for concurrency to prevent excessive memory allocation
+	const maxAllowedConcurrency = 100
+	concurrency := a.MaxConcurrency
+	if concurrency <= 0 {
+		concurrency = 10 // Default value if not set
+	} else if concurrency > maxAllowedConcurrency {
+		concurrency = maxAllowedConcurrency // Cap at a reasonable maximum
+	}
+
+	semaphore := make(chan struct{}, concurrency)
 	var wg sync.WaitGroup
 	var mutex sync.Mutex
 	verbose := viper.GetBool("verbose")
--- a/pkg/analysis/analysis_test.go
+++ b/pkg/analysis/analysis_test.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/agiledragon/gomonkey/v2"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/ai"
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Verified both failure modes against the source: `RunCustomAnalysis` builds its semaphore from raw `a.MaxConcurrency` (analysis.go L263), and the dispatch does `semaphore <- struct{}{}` *before* launching the goroutine that drains it (L280–281) — so capacity 0 deadlocks the whole `analyze` command and a negative value panics in `make`. Meanwhile `RunAnalysis` already clamps (default 10, cap 100); this fix copies that exact clamp, so the two paths now behave identically, which is the right shape for the fix.

The tests are the right ones too: the zero-concurrency test guards against the deadlock with a timeout harness, and the negative test locks in no-panic.

Non-blocking: the clamp block is now duplicated verbatim in `RunAnalysis` and `RunCustomAnalysis` — a tiny `clampConcurrency(n int) int` helper would keep them from drifting. Fine as a follow-up.

CI is green. Approving — recommending to @AlexsJones for final look and merge.

PR #1701, opened 2026-07-13
fix: use empty default for auth add --model so per-backend fallback runs

anxkhn:
Closes # 

`k8sgpt auth add -b anthropic`, run without `--model`, persists `model: gpt-4o`
for the Anthropic provider. The Anthropic API rejects `gpt-4o`, so the backend
does not work out of the box.

Root cause is a dead-code guard introduced with the Anthropic backend (#1652).
`cmd/auth/add.go` picks a per-backend default model only when the model is empty:

```go
// cmd/auth/add.go
if model == "" {
    fallbackModel := defaultModel
    if backend == "anthropic" {
        fallbackModel = anthropicDefaultModel // "claude-3-5-sonnet-latest"
    }
    model = fallbackModel
}
```

But the `--model` flag was registered with a **non-empty** default:

```go
addCmd.Flags().StringVarP(&model, "model", "m", defaultModel, "Backend AI model")
// defaultModel = "gpt-4o"
```

So when `--model` is omitted, cobra binds `model` to `"gpt-4o"`, the `model == ""`
guard never runs, and the Anthropic branch is unreachable. The client-side safety
net in `pkg/ai/anthropic.go` (`if model == "" { model = anthropicDefaultModel }`)
is also bypassed, because the persisted config already holds a non-empty `gpt-4o`.

### Fix

Register `--model` with an empty default, matching how `auth update` already does
it (`cmd/auth/update.go`). This lets the existing fallback resolve the correct
default per backend:

- `openai`  -> `gpt-4o`
- `anthropic` -> `claude-3-5-sonnet-latest`

`defaultModel` is still used as the OpenAI fallback inside the guard, so nothing
becomes unused.

```diff
-	addCmd.Flags().StringVarP(&model, "model", "m", defaultModel, "Backend AI model")
+	addCmd.Flags().StringVarP(&model, "model", "m", "", "Backend AI model")
```

Behavior when `--model` is passed explicitly is unchanged.

### Test

Adds `TestAddCommandModelDefaultPerBackend`, a table-driven test (anthropic +
openai) in `cmd/auth/provider_helpers_test.go`. It seeds `model` from the flag's
registered default to reproduce a real `auth add` without `--model`, then asserts
the persisted provider model. With the old non-empty default the anthropic case
fails (`got "gpt-4o"`); with the fix both cases pass. No cluster or API key
required.

## ✅ Checks

## ℹ Additional Information
No breaking changes. Explicit `--model` values are unaffected; only the
omitted-`--model` default path changes, from always `gpt-4o` to the correct
per-backend default. Related prior art: #1603 (open) refactors this same block
into a `defaultModelForBackend` helper but keeps the unreachable guard; this fix
targets the flag registration (a different line), so the two do not conflict and
this change would make that helper reachable too.

### DIFF
--- a/cmd/auth/add.go
+++ b/cmd/auth/add.go
@@ -165,7 +165,7 @@ func init() {
 	// add flag for backend
 	addCmd.Flags().StringVarP(&backend, "backend", "b", defaultBackend, "Backend AI provider")
 	// add flag for model
-	addCmd.Flags().StringVarP(&model, "model", "m", defaultModel, "Backend AI model")
+	addCmd.Flags().StringVarP(&model, "model", "m", "", "Backend AI model")
 	// add flag for password
 	addCmd.Flags().StringVarP(&password, "password", "p", "", "Backend AI password")
 	// add flag for url
--- a/cmd/auth/add.go
+++ b/cmd/auth/add.go
@@ -168,7 +168,7 @@ func init() {
 	// add flag for backend
 	addCmd.Flags().StringVarP(&backend, "backend", "b", defaultBackend, "Backend AI provider")
 	// add flag for model
-	addCmd.Flags().StringVarP(&model, "model", "m", defaultModel, "Backend AI model")
+	addCmd.Flags().StringVarP(&model, "model", "m", "", "Backend AI model")
 	// add flag for password
 	addCmd.Flags().StringVarP(&password, "password", "p", "", "Backend AI password")
 	// add flag for url
--- a/cmd/auth/provider_helpers_test.go
+++ b/cmd/auth/provider_helpers_test.go
@@ -141,6 +141,59 @@ func TestApplyAIProviderUpdatesIncludesAzureAPIVersion(t *testing.T) {
 	}
 }
 
+func TestAddCommandModelDefaultPerBackend(t *testing.T) {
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Verified the dead-code claim: the `--model` flag registers with `defaultModel` ("gpt-4o") as its default, so `model == ""` was unreachable when the flag was omitted and the per-backend fallback (including `anthropicDefaultModel`) never ran — `auth add -b anthropic` really does persist `gpt-4o` today. Changing the registered default to `""` makes the guard reachable and is also consistent with `auth update`, which already registers the same shared `model` var with an empty default.

Two side effects checked, both fine:
- `auth add` without `--model` now prints the existing yellow "model input is empty, will use the default value: X" warning on every run — slightly noisy but accurate, and it tells the user which default they got.
- `--help` no longer advertises `gpt-4o` as the default; given the default is now backend-dependent, that's more honest, though a help-text note like "(default depends on backend)" would be a nice touch.

The table-driven test exercises the real cobra binding path (seeding from the flag's `DefValue`) for both backends — exactly the regression lock this needs.

CI is green. Approving — recommending to @AlexsJones for final look and merge.

anxkhn:

PR #1702, opened 2026-07-13
fix: guard against nil backend port in HTTPRoute analyzer

anxkhn:
Closes #

The HTTPRoute analyzer panics with a nil pointer dereference when a `backendRef`
omits its optional `port`.

While validating that each `backendRef` targets an existing Service on a matching
port, the analyzer dereferences `*backend.Port`:

```go
for _, svcPort := range service.Spec.Ports {
    if int32(*backend.Port) == svcPort.Port {
        portMatch = true
    }
}
```

`Port` is optional on a `BackendObjectReference` (`*gatewayv1.PortNumber`, marked
`+optional`), so it is `nil` whenever a backendRef leaves it unset, which is valid
and common. A single portless backendRef makes the whole `k8sgpt analyze` run
crash with a nil pointer dereference.

This adds an early `if backend.Port == nil { continue }` before the port-match
loop, so a backendRef without a port is skipped for the port check instead of
being dereferenced. That is consistent with how the other optional pointer
accesses in the Gateway analyzers are already guarded.

The change is a 5-line guard in `pkg/analyzer/httproute.go` plus a regression
test in `pkg/analyzer/httproute_test.go` (`TestSvcNilPortHTTRouteAnalyzer`) that
builds an HTTPRoute whose backendRef has a nil `Port` and asserts `Analyze`
returns without panicking. Reverting only the source guard makes the new test
panic, confirming it is a genuine regression test.

## Checks

## Additional Information

Same crash class as the previously merged Service `TargetRef` nil-guard
(#1672); this applies the same defensive fix to the HTTPRoute backendRef port
path. No behavior change for backendRefs that do set a port. No dependency or
API changes.

### DIFF
--- a/pkg/analyzer/httproute.go
+++ b/pkg/analyzer/httproute.go
@@ -177,6 +177,11 @@ func (HTTPRouteAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 						},
 					})
 				} else {
+					// Port is optional on a backendRef; skip the port check when
+					// it is unset rather than dereferencing a nil pointer.
+					if backend.Port == nil {
+						continue
+					}
 					portMatch := false
 					for _, svcPort := range service.Spec.Ports {
 						if int32(*backend.Port) == svcPort.Port {
--- a/pkg/analyzer/httproute_test.go
+++ b/pkg/analyzer/httproute_test.go
@@ -207,6 +207,73 @@ func TestGWConfigSameHTTRouteAnalyzer(t *testing.T) {
 		t.Errorf("Expected message, <%s> , not found in HTTPRoute's analysis results", want)
 	}
 }
+
+func TestSvcNilPortHTTRouteAnalyzer(t *testing.T) {
+	// Service the backendRef points to, so analysis reaches the port-match loop.
+	Service := corev1.Service{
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Reviewed the diff — correct, minimal fix to a real crash. `BackendObjectReference.Port` is `*gatewayv1.PortNumber` and the analyzer dereferenced it unconditionally inside the port-match loop, so any backendRef without a port panicked the entire analyze run. The guard short-circuits before the deref, and the regression test is built properly: the Service exists so analysis genuinely reaches the port-match branch (a missing Service would take the other arm and never exercise the fix), and `BuildHTTPRoute` already accepted a `*PortNumber`, so passing `nil` needed no helper surgery.

One non-blocking observation for a possible follow-up: the Gateway API spec says Port is *required* when the referent is a Kubernetes Service — older CRDs just don't enforce it. So a nil port on a Service backendRef is arguably itself a misconfiguration the analyzer could surface as a failure rather than silently skipping the port check. Skipping is the safe, correct behavior for this fix (crash → no crash, no false positives); flagging it is a separate product decision.

CI green. Approving — recommending to @AlexsJones for final look and merge.

anxkhn:

PR #1703, opened 2026-07-13
fix: mask real pod name and namespace in statefulset analyzer

anxkhn:
The StatefulSet analyzer leaks the real pod name to the LLM when `--anonymize`
is set, because two of its `Sensitive` entries reference the wrong fields.

In the "pod not running" failure, the text is built from `pod.Name` and
`pod.Namespace`:

```go
Text: fmt.Sprintf("Statefulset pod %s in the namespace %s is not in running state.", pod.Name, pod.Namespace),
```

but the `Sensitive` entries paired the wrong `Unmasked` values with the masks:

```go
{ Unmasked: sts.Namespace, Masked: util.MaskString(pod.Name) },
{ Unmasked: serviceName,   Masked: util.MaskString(pod.Namespace) },
```

Anonymization (`pkg/analysis/analysis.go`) replaces each `Unmasked` substring
found in the text via `util.ReplaceIfMatch(text, s.Unmasked, s.Masked)`. Since
neither `sts.Namespace` nor `serviceName` is the pod name that actually appears
in the text, the real pod name is never replaced and is sent to the LLM in the
clear; `serviceName` matches nothing at all. (The pod namespace only happens to
be masked when it equals the StatefulSet namespace, which is the common case,
so it is not reliably protected either.)

This sets each `Unmasked` to the field actually interpolated into the text, so
both the pod name and namespace are masked. This mirrors the already-correct
pattern used a few lines above in the same analyzer for the "service does not
exist" failure.

```go
{ Unmasked: pod.Name,      Masked: util.MaskString(pod.Name) },
{ Unmasked: pod.Namespace, Masked: util.MaskString(pod.Namespace) },
```

A regression test (`TestStatefulSetAnalyzerNotRunningPodSensitiveMasking`)
reproduces the analyzer output for a pending StatefulSet pod, applies the same
masking loop the analysis code uses, and asserts the raw pod name and namespace
do not survive. It fails on the old code (pod name leaks) and passes with the fix.

## ✅ Checks

## ℹ Additional Information

No behavior change to analyzer output or messages; only the values masked under
`--anonymize` change. No API, dependency, or documentation changes.

### DIFF
--- a/pkg/analyzer/statefulset.go
+++ b/pkg/analyzer/statefulset.go
@@ -116,11 +116,11 @@ func (StatefulSetAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 						Text: fmt.Sprintf("Statefulset pod %s in the namespace %s is not in running state.", pod.Name, pod.Namespace),
 						Sensitive: []common.Sensitive{
 							{
-								Unmasked: sts.Namespace,
+								Unmasked: pod.Name,
 								Masked:   util.MaskString(pod.Name),
 							},
 							{
-								Unmasked: serviceName,
+								Unmasked: pod.Namespace,
 								Masked:   util.MaskString(pod.Namespace),
 							},
 						},
--- a/pkg/analyzer/statefulset_test.go
+++ b/pkg/analyzer/statefulset_test.go
@@ -15,10 +15,12 @@ package analyzer
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
+	"github.com/k8sgpt-ai/k8sgpt/pkg/util"
 	"github.com/magiconair/properties/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Reviewed the diff and re-derived the leak against `main` — the claim holds. The failure text is built from `pod.Name` and `pod.Namespace`, but the `Sensitive` pairs listed `sts.Namespace` and `serviceName` as the unmasked values. Since anonymization works by substring replacement (`util.ReplaceIfMatch(text, s.Unmasked, s.Masked)` in `pkg/analysis`), a value that never appears in the text is simply never replaced: the service name matches nothing, and the real pod name — the most identifying token in the sentence — sailed through to the LLM with `--anonymize` set. The fix pairs each mask with the value that actually appears in the text, which is the entire contract of `Sensitive`.

The test is the right kind for a privacy bug: rather than asserting struct fields, it runs the failure through the same `ReplaceIfMatch` pipeline the anonymizer uses and asserts the raw pod name and namespace are gone from the output — so it fails against `main` for the real reason and locks the masking end-to-end. Fixture drives the exact branch (available replicas 0, pod Pending, backing Service present).

Worth flagging for the broader theme: this is the same *"Sensitive entry doesn't match the text"* class of bug as the anonymization issues fixed before, and it's invisible to any test that only checks `Text`. A sweep of the other analyzers' `Sensitive` pairs against their `fmt.Sprintf` arguments would probably find siblings — good follow-up material.

CI green. Approving — recommending to @AlexsJones for final look and merge.

anxkhn:

PR #1704, opened 2026-07-13
fix: return a fresh AI client per call to avoid shared-state data race

anxkhn:
```
Closes #<ISSUE>

## 📑 Description

`ai.NewClient` looked up the requested backend in a package-level slice of
pre-allocated client instances and returned that shared instance directly. Every
caller then configures it in place via `Configure` (for example
`OpenAIClient.Configure` sets `client`, `model`, `temperature` and `topP`).

Under the CLI this is harmless because analysis runs sequentially. In server mode
it is not: each gRPC/MCP Analyze request builds its own `analysis.Analysis`, and
grpc-go serves every RPC in its own goroutine with no serializing lock. Concurrent
requests for the same backend therefore race on the singleton's fields, and one
request can overwrite another's configuration (model, temperature, base URL, token,
custom headers), so a request may end up calling the model with another request's
settings.

This makes the `clients` registry a list of constructors and has `NewClient` invoke
the matching constructor, so every call returns a distinct client and each analysis
owns its own. Callers already treat the return value as their own, so no other code
changes are needed. `NewClient` still returns `IAI` and the default fallback still
returns a fresh `&OpenAIClient{}`, so the change is source-compatible.

## ✅ Checks

## ℹ Additional Information

Adds two regression tests in `pkg/ai/iai_test.go`:
`TestNewClientReturnsFreshInstance` (distinct instance per call) and
`TestNewClientConcurrentConfigure` (concurrent `Configure` calls, which trip the
race detector before this change and pass after).

Verified locally: `gofmt -l`, `go vet ./pkg/ai/...`, `go build ./...`,
`go test -race ./pkg/ai/`, and `go test ./pkg/analysis/... ./pkg/server/...` all
pass.
```

---

## Companion issue (open this FIRST, then put its number into `Closes #<ISSUE>`)

CONTRIBUTING asks contributors to discuss/track the work on an issue, and the PR
template wants `Closes #<id>`. There is no existing issue for this race, so open a
short bug report (same pattern as the merged nil-TargetRef fix, which paired a
2-line issue with the PR). If Anas would rather file PR-only, drop the `Closes`
line and paste the Description into the issueless body.

**Title:**
```
AI client is a shared singleton: concurrent server-mode Analyze requests race on its config
```

**Body:**
```
`ai.NewClient` (pkg/ai/iai.go) returns a shared, package-level client instance for
each backend rather than a fresh one. Callers immediately mutate that instance via
`Configure` (e.g. `OpenAIClient.Configure` sets client/model/temperature/topP).

In CLI use this is harmless because analysis is sequential. In server mode
(`k8sgpt serve`) each gRPC/MCP Analyze request builds its own analysis and grpc-go
runs each RPC in its own goroutine with no serializing lock, so concurrent requests
for the same backend race on the shared instance's fields and can overwrite one
another's configuration (model, temperature, base URL, token, custom headers). This
is both a data race (observable under `go test -race`) and a cross-request config
bleed.

Expected: `NewClient` should hand back a distinct client per call so each analysis
owns its own configuration.
```

### DIFF
--- a/pkg/ai/iai.go
+++ b/pkg/ai/iai.go
@@ -19,24 +19,28 @@ import (
 )
 
 var (
-	clients = []IAI{
-		&OpenAIClient{},
-		&AnthropicClient{},
-		&AzureAIClient{},
-		&LocalAIClient{},
-		&OllamaClient{},
-		&NoOpAIClient{},
-		&CohereClient{},
-		&AmazonBedRockClient{},
-		&AmazonBedrockConverseClient{},
-		&SageMakerAIClient{},
-		&GoogleGenAIClient{},
-		&HuggingfaceClient{},
-		&GoogleVertexAIClient{},
-		&OCIGenAIClient{},
-		&CustomRestClient{},
-		&IBMWatsonxAIClient{},
-		&GroqClient{},
+	// clients maps each backend to a constructor that returns a fresh client.
+	// NewClient must hand back a new instance per call so that concurrent
+	// callers (e.g. server-mode Analyze requests) do not share and race on a
+	// single client's fields via Configure.
+	clients = []func() IAI{
+		func() IAI { return &OpenAIClient{} },
+		func() IAI { return &AnthropicClient{} },
+		func() IAI { return &AzureAIClient{} },
+		func() IAI { return &LocalAIClient{} },
+		func() IAI { return &OllamaClient{} },
+		func() IAI { return &NoOpAIClient{} },
+		func() IAI { return &CohereClient{} },
+		func() IAI { return &AmazonBedRockClient{} },
+		func() IAI { return &AmazonBedrockConverseClient{} },
+		func() IAI { return &SageMakerAIClient{} },
+		func() IAI { return &GoogleGenAIClient{} },
+		func() IAI { return &HuggingfaceClient{} },
+		func() IAI { return &GoogleVertexAIClient{} },
+		func() IAI { return &OCIGenAIClient{} },
+		func() IAI { return &CustomRestClient{} },
+		func() IAI { return &IBMWatsonxAIClient{} },
+		func() IAI { return &GroqClient{} },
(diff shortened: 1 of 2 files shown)


AlexsJones (commented):
Thanks @anxkhn — the diagnosis is exactly right (the package-level `clients` slice of shared instances + in-place `Configure` is a real data race under server-mode concurrency), and the write-up is clear.

However, this collides head-on with **#1691** (@glatinone), which fixes the same bug in the same file and was reviewed-and-approved on 13 Jul, a few hours before this PR was filed. Both convert the shared registry into per-call constructors; #1691 uses a `map[string]func() IAI` keyed by backend name, which has two small advantages over the slice approach here: lookup doesn't allocate an instance of every registered backend on the way to a match (`NewClient` here constructs up to 17 clients to return one), and it drops the `GetName()` round-trip entirely.

So rather than request changes on working code, I'm flagging the overlap for @AlexsJones to pick a consolidation path — the likely outcome is #1691 lands (it was first and its registry shape is a bit stronger) and this one closes as superseded. Your `TestNewClientConcurrentConfigure` exercising `Configure` under `-race` is a genuinely useful addition that #1691's tests don't fully cover — if #1691 lands first, that test would be very welcome as a small follow-up PR, and your other fixes in this series (#1702, #1703 already approved) stand on their own.

AlexsJones:
Thanks again @anxkhn — closing this one, but not for lack of quality. Your diagnosis of the shared-client race was exactly right; it's the same bug #1691 (filed a little earlier) fixes with a map-by-name constructor registry, and Alex has decided to land that shape. This isn't a knock on your work: your other fixes (#1702, #1703) are approved on their own merits, and your concurrent `Configure` `-race` test here covers something #1691's tests don't — we'd genuinely welcome it as a small follow-up PR on top of the merged registry. Hope to see it!

Issue #1063, opened 2024-04-13
[bug]:  run k8sgpt analyze will only complete successfully once , after active trivy integration


liyuerich:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

0.3.29 (5db4bc2)

### Kubernetes Version

v1.27.5

### Host OS and its Version

Ubuntu Linux controller-node-1 5.4.0-174-generic

### Steps to reproduce

1. first I active trivy and run k8sgpt analyze successfully, 
2. then I run k8sgpt analyze again, I got error message.
3. after deactive trivy, run k8sgpt analyze again. it completed successfully.

error message:
 k8sgpt analyze --explain
fatal error: concurrent map writes
goroutine 16 [running]:
[k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName(0xc00014b3b0](http://k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName(0xc00014b3b0), {{0x2be7d26, 0x16}, {0x2bc5c63, 0x8}, {0x241e398, 0x15}}, {0x345f568?, 0xc0009b4310})
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/runtime/scheme.go:181 +0x345
[k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypes(0xc00014b3b0](http://k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypes(0xc00014b3b0), {{0x2be7d26?, 0x0?}, {0x2bc5c63?, 0x0?}}, {0xc000854620?, 0x16, 0x0?})
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/runtime/scheme.go:148 +0x176
[github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1.addKnownTypes(0xc0008547b8?)](http://github.com/aquasecurity/trivy-operator/pkg/apis/aquasecurity/v1alpha1.addKnownTypes(0xc0008547b8?))
[...]

PR #1705, opened 2026-07-13
fix: register gateway-api scheme once to avoid concurrent map writes

anxkhn:
Refs #1063 (gateway-api facet; see note below)

The `Gateway`, `GatewayClass` and `HTTPRoute` analyzers each call
`gtwapi.AddToScheme(client.Scheme())` at the top of `Analyze()`:

```go
client := a.Client.CtrlClient
err := gtwapi.AddToScheme(client.Scheme())
if err != nil {
    return nil, err
}
```

`RunAnalysis` dispatches every analyzer in its own goroutine
(`go a.executeAnalyzer(...)`) against a single shared `CtrlClient`. The three
gateway-api analyzers therefore register the gateway-api types into the same
`runtime.Scheme` concurrently. `runtime.Scheme.AddKnownTypeWithName` writes its
internal maps without any lock, so this is a data race that Go turns into a hard
`fatal error: concurrent map writes`, taking down the whole process. In practice
`k8sgpt analyze` only completes reliably when a single run happens to serialize
these registrations, which is why users report it "only completes successfully
once" and then crashes on subsequent runs.

Fix: register the gateway-api types once, when the client is built in
`NewClient`, off the concurrent hot path, and drop the three per-`Analyze`
`AddToScheme` calls. The types are on the shared scheme before any analyzer runs,
so the goroutines only read the scheme and no longer race on it.

`gtwapi.Install` is used rather than the deprecated `gtwapi.AddToScheme`
(the latter is annotated `// Deprecated: use Install instead`); both resolve to
the same scheme builder, and `Install` is already the idiom used in the existing
gateway analyzer tests.

## Testing

Added `pkg/analyzer/concurrent_scheme_test.go`, which runs the three gateway-api
analyzers in parallel (50 iterations each) against one shared client, mirroring
how `RunAnalysis` fans out. It fails under `-race` on the old code with the
`concurrent map writes` / `DATA RACE` stack from #1063 and passes on this change.

Commands run locally (all pass):

```
gofmt -l <changed files>            # empty
go vet ./pkg/analyzer/... ./pkg/kubernetes/...
go build ./...
go test ./pkg/analyzer/... -race
go test ./pkg/kubernetes/... -race
go test ./...
```

## ✅ Checks

## ℹ Additional Information

#1063 is an umbrella "concurrent map writes" report that covers several distinct
analyzers. This PR fixes only the gateway-api facet (the `GatewayClassAnalyzer.
Analyze -> gateway-api addKnownTypes` stack shown in the issue), so it uses
`Refs #1063` rather than a closing keyword to avoid auto-closing the umbrella
while other facets remain. The trivy facet is already handled by the open
#1222; the external-secrets and kyverno facets are out of scope here and left
for follow-up PRs to keep this change small and focused.

### DIFF
--- a/pkg/analyzer/concurrent_scheme_test.go
+++ b/pkg/analyzer/concurrent_scheme_test.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package analyzer
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
+	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	gtwapi "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// TestGatewayAnalyzersConcurrentScheme guards against issue #1063: running the
+// gateway-api analyzers in parallel used to crash with "fatal error:
+// concurrent map writes" because each analyzer registered the gateway-api types
+// into the shared client scheme from inside Analyze(). The types are now
+// registered once when the client is built, so the analyzers can safely share a
+// single client across goroutines. Run with -race to catch a regression.
+func TestGatewayAnalyzersConcurrentScheme(t *testing.T) {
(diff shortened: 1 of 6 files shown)


AlexsJones (approved):
Reviewed the full diff. This is a correct fix for the #1063 crash class: `RunAnalysis` dispatches analyzers in goroutines against one shared `CtrlClient`, and the three gateway-api analyzers each called `gtwapi.AddToScheme(client.Scheme())` inside `Analyze()` — `runtime.Scheme`'s internal maps are unsynchronized, so concurrent registration is a `fatal error: concurrent map writes` that kills the whole process (not even a recoverable panic). Moving registration to `kubernetes.NewClient`, once, off the hot path, is the right shape, and removing all three in-`Analyze()` call sites means no stragglers.

The regression test is real: 150 concurrent `Analyze` calls across the three analyzers under `-race` would reliably catch a reintroduction. Two non-blocking notes:

- The test installs gateway-api into the global `scheme.Scheme` from client-go rather than a fresh `runtime.NewScheme()` — that mutates package-global state other tests in the package share. Harmless today, but a fresh scheme would be more hygienic.
- Any code path that builds a `CtrlClient` without going through `kubernetes.NewClient` now gets "no kind registered" errors from the gateway analyzers instead of registering lazily. I checked the obvious paths and they all route through `NewClient`, and that failure mode is a clean error rather than a crash — acceptable.

CI green (lint, build, validate). Approving — recommending to @AlexsJones for final look and merge.

anxkhn:

PR #1706, opened 2026-07-14
fix: report keda scaledobject failures when no event exists

anxkhn:
Closes #

The KEDA `ScaledObject` analyzer collects failures for a scaled object (for
example `"<Kind> <ns>/<name> does not have resource configured."` when a
cpu/memory-triggered target declares no resource requests/limits, and the
"ScaleTargetRef which does not exist" diagnostic), then calls
`util.FetchLatestEvent` and did an unconditional `continue` whenever that
returned a nil event:

```go
evt, err := util.FetchLatestEvent(a.Context, a.Client, so.Namespace, so.Name)
if err != nil || evt == nil {
    continue
}
```

`util.FetchLatestEvent` returns `(nil, nil)` for any object that has no recent
event, which is the common case. Because that `continue` lives in the outer
`for _, so := range list.Items` loop, it skips the shared
`if len(failures) > 0 { ... }` commit block at the end of the iteration, so every
failure already collected for that `ScaledObject` is silently discarded. In
practice, a genuinely misconfigured `ScaledObject` reports nothing unless it also
happens to have a recent non-Normal event.

The fix drops the early `continue` and only appends the event-based failure when
an event is actually present and not `Normal`, so control always reaches the
commit block:

```go
evt, err := util.FetchLatestEvent(a.Context, a.Client, so.Namespace, so.Name)
if err == nil && evt != nil && evt.Type != "Normal" {
    failures = append(failures, common.Failure{ ... })
}
```

This matches the sibling analyzers (for example `pkg/analyzer/statefulset.go`),
which call `FetchLatestEvent` and fall through to record previously collected
failures rather than `continue`-ing past the commit block.

A colocated `scaledobject_analyzer_test.go` is added. The analyzer builds its
KEDA typed client from `rest.Config`, so the test serves the `ScaledObject` list
from an `httptest` server and backs the target `Deployment`/events with the
core `client-go` fake. It covers:

- the regression: a misconfigured target with no event now reports
  "does not have resource configured." (previously reported nothing);
- a missing `ScaleTargetRef` with no event still reports;
- a `Warning` event is still surfaced.

## Checks

## Additional Information

Behavior-only change scoped to `ScaledObjectAnalyzer.Analyze`; no dependency or
public-API changes (`go.mod`/`go.sum` unchanged). Introduced by the original
KEDA integration (#1058) and not previously reported.


AlexsJones (approved):
The claimed bug is real — verified against main rather than taking the PR's word for it. In `pkg/integration/keda/scaledobject_analyzer.go` on main, the event check sits in the outer `for _, so := range list.Items` loop:

```go
evt, err := util.FetchLatestEvent(a.Context, a.Client, so.Namespace, so.Name)
if err != nil || evt == nil {
    continue
}
```

`util.FetchLatestEvent` returns `(nil, nil)` whenever no event matches `involvedObject.name=<name>` — the common case — so that `continue` bypasses the `if len(failures) > 0 { ... }` commit block at the bottom of the iteration. Any "does not have resource configured." failure collected earlier for that ScaledObject is silently discarded: a misconfigured cpu/memory-triggered ScaledObject with no recent event reports nothing on main.

**The fix.** Collapsing to `if err == nil && evt != nil && evt.Type != "Normal"` is the right shape: control always reaches the commit block, and the event-based failure is only appended when a non-Normal event actually exists. Side effects checked:

- *Healthy objects:* unchanged — with no failures collected, the commit block is a no-op whether or not an event exists.
- *Error path:* a `FetchLatestEvent` error no longer discards previously collected failures — it now skips only the event-based failure, matching how `pkg/analyzer/statefulset.go` scopes the same call. Consistent with repo convention.

**Tests.** Real, not tautological, with one caveat:

- `TestScaledObjectAnalyzerReportsResourceFailureWithoutEvent` is a genuine regression test — on main the nil event triggers the `continue` and `assert.Len(t, results, 1)` fails. The httptest-server approach for the KEDA typed client is necessary (the analyzer builds it from `rest.Config` via `v1alpha1.NewForConfig`, so a fake clientset can't intercept it) and is done cleanly.
- `TestScaledObjectAnalyzerReportsWarningEvent` correctly pins the preserved event path.
- `TestScaledObjectAnalyzerReportsMissingTargetWithoutEvent` is valid coverage but **not** a regression test, and its comment ("another path that the previous early-continue discarded") is inaccurate — as is the same claim in the PR body. On main, the `continue` lives inside the `else` branch that only executes when `podInfo != nil`; the `ScaleTargetRef which does not exist` failure (the `podInfo == nil` branch) never reached the `continue` and was already reported correctly. Non-blocking, but worth correcting the comment (pre- or post-merge) so future readers don't misread the bug's actual scope.

Also checked: the missing-target test calls `GetApiDocV2` with a nil `OpenapiSchema`; gnostic's generated getters are nil-safe, so no panic — consistent with green CI.

**CI** fully green: golangci-lint, build/tests, Semantic PR, DCO, EasyCLA, License Compliance, both codecov checks; multi-arch image build skipped as expected for PRs.

Approving — recommending to @AlexsJones for merge, optionally with the one comment correction.

anxkhn:

PR #1707, opened 2026-07-14
fix: populate ParentObject in ClusterCatalog and ClusterExtension analyzers

anxkhn:
Closes # 

## 📑 Description

The `ClusterCatalog` and `ClusterExtension` OLM analyzers never populate
`ParentObject` on their results.

Both analyzers store the object they are analyzing in `value.Catalog` /
`value.Extension`, but they look up the owning resource with:

```go
parent, found := util.GetParent(a.Client, value.Node.ObjectMeta)
```

`value.Node` is never set by these analyzers, so `GetParent` always ran against a
zero-value `ObjectMeta` (nil `OwnerReferences`), returned `("", false)`, and
`ParentObject` stayed empty. This is a copy-paste leftover from a Node-based
analyzer.

The fix reads the `ObjectMeta` from the field each analyzer actually stores,
matching every other analyzer in the package:

- `pkg/analyzer/clustercatalog.go`: `value.Node.ObjectMeta` -> `value.Catalog.ObjectMeta`
- `pkg/analyzer/clusterextension.go`: `value.Node.ObjectMeta` -> `value.Extension.ObjectMeta`

Each change also adds a focused regression test: a CR with an `ownerReferences`
entry pointing at a Deployment (seeded in the fake typed client) now asserts that
`ParentObject == "Deployment/<owner>"`. Both tests fail (`ParentObject == ""`)
without the one-line fix and pass with it.

This is the same class of fix as the previously merged #144 (`fix: ingress object
meta value`) and #1068 (`fix: invalid ParentObj in output`).

## ✅ Checks

## ℹ Additional Information

No behavior change beyond correctly resolving the parent object for these two
analyzers; no new dependencies. Verified locally:

- `gofmt -l` on the 4 changed files: clean
- `go vet ./pkg/analyzer/...`: clean
- `go build -o k8sgpt .`: ok
- `go test ./pkg/analyzer/... -count=1`: ok
- targeted: `go test ./pkg/analyzer/ -run 'TestClusterCatalogAnalyzerParentObject|TestClusterExtensionAnalyzerParentObject' -v`: both PASS

### DIFF
--- a/pkg/analyzer/clustercatalog.go
+++ b/pkg/analyzer/clustercatalog.go
@@ -82,7 +82,7 @@ func (ClusterCatalogAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error
 			Error: value.FailureDetails,
 		}
 
-		parent, found := util.GetParent(a.Client, value.Node.ObjectMeta)
+		parent, found := util.GetParent(a.Client, value.Catalog.ObjectMeta)
 		if found {
 			currentAnalysis.ParentObject = parent
 		}
--- a/pkg/analyzer/clustercatalog_test.go
+++ b/pkg/analyzer/clustercatalog_test.go
@@ -21,6 +21,8 @@ import (
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
 	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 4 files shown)


AlexsJones (approved):
**Bug confirmed on main.** Both `pkg/analyzer/clustercatalog.go` and `pkg/analyzer/clusterextension.go` build their `PreAnalysis` entries via `Catalog: *catalog` / `Extension: *extension`, but then call `util.GetParent(a.Client, value.Node.ObjectMeta)`. `common.PreAnalysis.Node` is never set by these analyzers, so `GetParent` receives a zero `ObjectMeta`; since `meta.OwnerReferences == nil` it returns `("", false)` immediately, and `ParentObject` was unconditionally empty. Classic copy-paste from the Node analyzer, as the PR description says.

**Fix is correct and consistent.** `common.ClusterCatalog` and `common.ClusterExtension` both embed `metav1.ObjectMeta` (populated by `runtime.DefaultUnstructuredConverter` in `ConvertToClusterCatalog`/`ConvertToClusterExtension`), so `value.Catalog.ObjectMeta` / `value.Extension.ObjectMeta` carry the real `ownerReferences`. This matches the established pattern in sibling analyzers (e.g. `pkg/analyzer/ingress.go`: `util.GetParent(a.Client, value.Ingress.ObjectMeta)`).

**User-visible impact.** `ParentObject` is printed next to each result in text output (`pkg/analysis/output.go`) and serialized as `parentObject` in JSON, so affected results previously showed an empty parent. Practical impact is modest — these cluster-scoped OLMv1 CRs rarely carry owner references to the workload kinds `GetParent` resolves — but the code is now correct rather than dead, and aligned with every other analyzer.

**Tests are real regression tests.** Each seeds a CR whose `ownerReferences` point at a Deployment in the fake typed clientset, drives a genuine validation failure (Progressing/Retrying condition for the catalog; invalid `upgradeConstraintPolicy` for the extension), then asserts `results[0].ParentObject == "Deployment/<owner>"`. Revert the one-line fix and the assertion fails against `""`.

**CI** fully green: golangci-lint, build/tests, Semantic PR, DCO, EasyCLA, License Compliance, both codecov checks; multi-arch image push skipped as expected for PRs.

Nits (non-blocking): the PR body's `Closes #` line is empty template leftover; `Namespace: "test"` on the `common.Analyzer` config in both new tests is inert for cluster-scoped analyzers — harmless.

Approving — recommending to @AlexsJones for merge.

anxkhn:

PR #1708, opened 2026-07-15
fix: prevent nil prometheus config dereference on invalid gzipped config

anxkhn:
`unmarshalPromConfigBytes` in `pkg/integration/prometheus/config_analyzer.go`
could return a `nil` `*promconfig.Config` together with an error on one path:
when the input is detected as gzip (`http.DetectContentType(b) == "application/x-gzip"`),
gunzips successfully, but the inner `yaml.Unmarshal` fails. That single branch
returned `nil, err` while every other branch of the same function returns
`&config`.

Both callers then dereference the returned config without a nil check, so this
input crashes the process instead of surfacing a validation failure:

- `RelabelAnalyzer.Analyze` discards the error (`config, _ := unmarshalPromConfigBytes(pc.b)`)
  and immediately ranges `config.ScrapeConfigs`, panicking on the nil pointer.
- `ConfigAnalyzer.Analyze` records the error but does not `continue`, then reads
  `len(config.ScrapeConfigs)`, hitting the same nil-pointer dereference.

Since the analyzers deliberately route malformed configurations through this
helper to report validation failures, a corrupt or truncated gzip payload in a
Prometheus `ConfigMap`/`Secret` is a realistic trigger, and a hard panic is not
the intended outcome.

The fix makes the helper total: return the (empty) `&config` alongside the error
on that path, matching every other branch. Both analyzers then behave as
intended: `ConfigAnalyzer` surfaces the unmarshal error plus the empty
scrape-config failure, and `RelabelAnalyzer` ranges an empty slice without
panicking. Only the previously-panicking malformed-config path changes behavior.

## Checks

## Additional Information

- One-line production change (`config_analyzer.go`, the gunzip/invalid-YAML
  branch) plus a colocated `config_analyzer_test.go`.
- New tests:
  - `TestUnmarshalPromConfigBytesGzipInvalid` asserts the helper never returns a
    nil `*Config` on the gzip-detected/invalid-YAML path (the previously
    panicking dereference is now safe).
  - `TestPrometheusAnalyzersGzipInvalidConfig` drives both analyzers end to end
    with a fake clientset (a Prometheus pod plus a `ConfigMap` holding the
    gzip-detected invalid payload) and asserts they report failures without
    panicking.
- Verified red -> green: reverting the one-line change makes the helper test fail
  and `ConfigAnalyzer` panic with a nil-pointer dereference; with the change all
  tests pass.
- No breaking changes; no dependencies added.

### DIFF
--- a/pkg/integration/prometheus/config_analyzer.go
+++ b/pkg/integration/prometheus/config_analyzer.go
@@ -281,7 +281,7 @@ func unmarshalPromConfigBytes(b []byte) (*promconfig.Config, error) {
 		}
 		err = yaml.Unmarshal(gunzipBytes, &config)
 		if err != nil {
-			return nil, err
+			return &config, err
 		}
 		return &config, nil
 	} else {
--- a/pkg/integration/prometheus/config_analyzer_test.go
+++ b/pkg/integration/prometheus/config_analyzer_test.go
@@ -1,0 +1,138 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prometheus
+
+import (
+	"bytes"
+	"compress/gzip"
+	"context"
+	"net/http"
+	"testing"
+
+	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
+	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Verified against `main`: the gzip-detected → gunzip-ok → inner `yaml.Unmarshal` failure branch is the only path in `unmarshalPromConfigBytes` that returns `(nil, err)` — every other branch returns `&config`, and both callers rely on that contract (`RelabelAnalyzer.Analyze` discards the error outright at `relabel_analyzer.go:34` and ranges over `config.ScrapeConfigs`; `ConfigAnalyzer.Analyze` records the failure but still reads `config.ScrapeConfigs`). So a gzip-content-typed blob that gunzips to invalid YAML panics both analyzers today, and the one-line fix restores the invariant the rest of the function already keeps.

The tests earn their keep: the helper-level test pins the "never return nil" contract with a `DetectContentType` guard so the fixture provably exercises the gunzip path, and the end-to-end test drives both analyzers over a fake clientset — ConfigAnalyzer surfaces the validation failure, RelabelAnalyzer stays quiet instead of crashing. CI is fully green including codecov on the new file.

Approving — ready for the maintainer's final look.

anxkhn:

PR #1712, opened 2026-07-21
fix: harden analyzers, AI providers, server, and cache against panics and hangs

Navneet072300:
## 📑 Description

A small sweep of bug fixes found while reading through the analyzer, AI provider, server, and cache packages. Each fix is an independent commit so they can be reviewed and reverted individually if needed.

- **`fix(analyzer)`**: the Ingress analyzer dereferenced `path.Backend.Service` unconditionally, panicking on any Ingress using a `resource` backend instead of a `service` backend (e.g. S3 or custom controller backends).
- **`fix(server)`**: the `Shutdown` gRPC handler was a stub that called `panic("implement me")`. Since reflection is enabled and the endpoint has no auth, any client could crash a running `k8sgpt serve` process with a single call. It now gracefully closes the listener instead.
- **`fix(ai)`**: the OpenAI, Azure OpenAI, and Groq clients indexed `resp.Choices[0]` without checking the slice was non-empty, panicking on a content-filtered/empty completion. Also added a default HTTP client timeout across the OpenAI-compatible, custom REST, Ollama, and Anthropic clients, since none had one and a hung endpoint could block `analyze --explain` indefinitely.
- **`fix(analysis)`**: `RunAnalysis` didn't validate that `--namespace` exists before running (unlike `RunCustomAnalysis`, which already does). A typo'd namespace silently reported "No problems detected" instead of an error.
- **`fix(analyzer)`**: the Deployment, DaemonSet, and Mutating/ValidatingWebhook analyzers called the Kubernetes API with `context.Background()` instead of the caller-supplied `a.Context`, unlike every other analyzer in the package.
- **`fix(cli)`**: corrected the `--with-stat` flag description, which claimed it disables error display — it doesn't; stats are printed in addition to normal output.
- **`fix(cache)`**: the file-based cache built file paths by joining the cache directory with a caller-supplied key without checking the result stayed inside that directory, allowing a key like `../../some-file` to resolve outside the cache subdirectory.

## ✅ Checks

## ℹ Additional Information
No behavior changes beyond the bug fixes described above. `go build ./...`, `go vet ./...`, and the full test suite pass locally. One existing test (`TestAnalysis_RunAnalysisWithFilter`) needed a `Namespace` fixture added to its fake clientset, since it started exercising the new namespace-validation path.

### DIFF
--- a/cmd/analyze/analyze.go
+++ b/cmd/analyze/analyze.go
@@ -176,5 +176,5 @@ func init() {
 	// label selector flag
 	AnalyzeCmd.Flags().StringVarP(&labelSelector, "selector", "L", "", "Label selector (label query) to filter on, supports '=', '==', and '!='. (e.g. -L key1=value1,key2=value2). Matching objects must satisfy all of the specified label constraints.")
 	// print stats
-	AnalyzeCmd.Flags().BoolVarP(&withStats, "with-stat", "s", false, "Print analysis stats. This option disables errors display.")
+	AnalyzeCmd.Flags().BoolVarP(&withStats, "with-stat", "s", false, "Print analysis stats (time taken per analyzer) in addition to the normal output.")
 }
--- a/pkg/ai/anthropic.go
+++ b/pkg/ai/anthropic.go
@@ -50,16 +50,16 @@ func (c *AnthropicClient) Configure(config IAIConfig) error {
 		opts = append(opts, option.WithAPIKey(token))
 	}
 
+	httpClient := &http.Client{Timeout: defaultHTTPTimeout}
 	proxyEndpoint := config.GetProxyEndpoint()
 	if proxyEndpoint != "" {
 		proxyURL, err := url.Parse(proxyEndpoint)
 		if err != nil {
 			return err
 		}
-		opts = append(opts, option.WithHTTPClient(&http.Client{
-			Transport: &http.Transport{Proxy: http.ProxyURL(proxyURL)},
-		}))
+		httpClient.Transport = &http.Transport{Proxy: http.ProxyURL(proxyURL)}
 	}
+	opts = append(opts, option.WithHTTPClient(httpClient))
 
 	model := config.GetModel()
 	if model == "" {
--- a/pkg/ai/azureopenai.go
+++ b/pkg/ai/azureopenai.go
@@ -58,6 +58,7 @@ func (c *AzureAIClient) Configure(config IAIConfig) error {
 			Proxy: http.ProxyFromEnvironment,
 		}
 		defaultConfig.HTTPClient = &http.Client{
+			Timeout: defaultHTTPTimeout,
 			Transport: &customHeaderRoundTripper{
 				headers: customHeaders,
 				rt:      transport,
@@ -73,6 +74,7 @@ func (c *AzureAIClient) Configure(config IAIConfig) error {
 		}
 
 		defaultConfig.HTTPClient = &http.Client{
+			Timeout:   defaultHTTPTimeout,
 			Transport: transport,
 		}
 	}
@@ -116,6 +118,9 @@ func (c *AzureAIClient) GetCompletion(ctx context.Context, prompt string) (strin
 	})
 	if err != nil {
 		return "", err
+	}
+	if len(resp.Choices) == 0 {
+		return "", errors.New("no completion choices returned")
 	}
 	return resp.Choices[0].Message.Content, nil
 }
--- a/pkg/ai/customrest.go
+++ b/pkg/ai/customrest.go
@@ -60,7 +60,7 @@ func (c *CustomRestClient) Configure(config IAIConfig) error {
 	c.base = baseClientURL
 
 	proxyEndpoint := config.GetProxyEndpoint()
-	c.client = http.DefaultClient
+	c.client = &http.Client{Timeout: defaultHTTPTimeout}
 	if proxyEndpoint != "" {
 		proxyUrl, err := url.Parse(proxyEndpoint)
 		if err != nil {
@@ -71,6 +71,7 @@ func (c *CustomRestClient) Configure(config IAIConfig) error {
 		}
 
 		c.client = &http.Client{
+			Timeout:   defaultHTTPTimeout,
 			Transport: transport,
 		}
 	}
--- a/pkg/ai/groq.go
+++ b/pkg/ai/groq.go
@@ -59,6 +59,7 @@ func (c *GroqClient) Configure(config IAIConfig) error {
 
 	customHeaders := config.GetCustomHeaders()
 	defaultConfig.HTTPClient = &http.Client{
+		Timeout: defaultHTTPTimeout,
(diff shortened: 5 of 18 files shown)


AlexsJones (commented):
Thanks @Navneet072300 — this is a careful, well-partitioned sweep, and every fix class checks out against `main`:

- **Ingress `resource` backend panic**: real. `path.Backend.Service` is nil for `Resource` backends and the current code dereferences it unconditionally — any Ingress pointing at, e.g., an S3/custom-controller backend panics the whole analyzer.
- **`Shutdown` RPC panic**: real and the nastiest of the set. gRPC-go does not recover handler panics by default, the endpoint has no auth, and reflection is on — a single `Shutdown` call kills a running `k8sgpt serve`. The `ShutdownFunc` injection keeps the listener lifecycle in the server package, which is the right shape.
- **`resp.Choices[0]` guards** (OpenAI/Azure/Groq): real — content-filtered or empty completions do return zero choices.
- **Default HTTP timeout**: the 10-minute value with the rationale comment in `iai.go` is sensible (generous enough for slow local Ollama, still bounded).
- **`context.Background()` → `a.Context`** in daemonset/deployment/webhook analyzers: good hygiene, consistent with the analyzers that already do this.
- **Cache path-traversal guard**: correct — `filepath.Join` cleans `..` before `xdg.CacheFile` sees it, so keys like `../foo` currently resolve outside the cache dir; the `baseDir` prefix check closes that.
- **`--with-stat` help text**: verified the old text was stale — stats print *in addition to* the normal output (`analyze.go` prints `output_data` unconditionally after `PrintStats`); nothing suppresses error display. The new wording matches actual behavior.

Two non-blocking notes:

1. The new namespace validation in `RunAnalysis` treats **any** error from `Namespaces().Get` as "namespace not found" and aborts the analysis. A user with namespace-scoped RBAC (can list pods in the namespace, cannot `get` the Namespace object itself) would get a hard failure where today the analysis works. It mirrors the existing check in `RunCustomAnalysis`, so consistency argues for keeping it as-is in this PR — but distinguishing `IsNotFound` (abort) from `Forbidden`/transient (warn and continue) would be a nice follow-up for both call sites.
2. `Shutdown` responding while a goroutine closes the listener has a small race (the response may not flush before the serve loop exits). In practice the client is tearing things down anyway; a `GracefulStop`-based path could be a future refinement, not something to hold this PR on.

One process item: the real CI (Run tests / Build container / golangci-lint) is sitting at **action_required** — first-time-contributor workflow approval — so nothing has actually built or run yet. @AlexsJones, worth approving the workflow runs; content-wise this is merge-ready once those come back green.

Navneet072300:
Hi @AlexsJones, is there anything you'd like me to change before this PR is merged? If everything looks good, when do you think it can be merged?

AlexsJones:
Nothing more needed from you, this is in the merge queue. The two pending checks are workflow runs that wait on a maintainer click for a first-time contributor, not anything in your changes. Thanks for your patience, the hardening set is solid work and the path-traversal cache fix in particular is appreciated.

AlexsJones:
The `Shutdown` handler find is a good one. `panic("implement me")` behind an unauthenticated, reflection-discoverable RPC means any client on the port can kill a running `k8sgpt serve`, and wiring it to the real listener lifecycle (async, so it doesn't close the socket out from under its own response) is the right shape. Same for the `context.Background()` to `a.Context` sweep in the webhook and deployment analyzers, those were quietly ignoring cancellation.

Three things before this can land:

**1. Please split it.** This is five unrelated fixes across analyzers, AI providers, the server and the cache. Convention in this repo is one narrow fix per PR (see #1725, #1728, #1729), and it matters here because these carry very different risk: the `Shutdown` fix should go in on its own today, whereas the namespace change below needs discussion. Separate commits help, but they don't let us merge the safe ones while arguing about the rest.

**2. The namespace pre-check in `RunAnalysis` is a regression risk.** Any `Get` failure is treated as "namespace not found" and aborts the whole run. A service account scoped to one namespace typically cannot `get` namespaces at all, so users who can analyze today would start getting `namespace "foo" not found` from a Forbidden. If you keep this, use `apierrors.IsNotFound(err)` to abort and let everything else fall through to the normal path. Worth a `--namespace` typo test either way, since that's the actual bug you're fixing.

**3. The Ingress `path.Backend.Service == nil` guard is already in #1728**, which is reviewed and queued to merge. Drop that hunk here so the two don't conflict.

Smaller one: the `--with-stat` help text now says stats print "in addition to the normal output", but nothing in this diff changes that behaviour. Is the old wording just stale, or is there a code change missing?

Navneet072300:
Thanks for the thorough re-verification — really appreciate you checking each fix class against `main` directly rather than taking the description at face value.

Splitting per your request, into six focused PRs:

- #1730 — `fix(server)`: Shutdown RPC (the safe one, ready to merge on its own)
- #1731 — `fix(ai)`: empty-choices guard + default HTTP timeout
- #1732 — `fix(analyzer)`: `context.Background()` → `a.Context` threading
- #1733 — `fix(cache)`: path traversal guard
- #1734 — `fix(cli)`: `--with-stat` help text
- #1735 — `fix(analysis)`: namespace validation, updated per your note below

On the two open items:

1. **Namespace check** — updated to use `apierrors.IsNotFound(err)` to abort, and everything else (including `Forbidden`) now falls through to the normal analysis path, exactly as you described. Added a test for both cases: a typo'd namespace still aborts with the "not found" error, and a `Forbidden` response from a namespace-scoped service account no longer blocks the analysis. See #1735.

2. **Ingress fix** — dropped from all of these; #1728 already covers it and is ahead in the queue, so no point in the two conflicting.

**On `--with-stat`**: confirmed it's stale wording, not a missing code change. `cmd/analyze/analyze.go` computes `output_data` via `PrintOutput` (which includes the Warnings block whenever `len(a.Errors) != 0`, unconditionally) and always prints it; when `--with-stat` is set, `PrintStats()` output is printed *before* it, as an addition, not a replacement. Nothing in the current code path (or this diff) suppresses errors. That's in #1734 with the corrected text.

Closing this one out in favor of the six above — let me know if you'd like anything further split or adjusted on any of them.

AlexsJones:
Thanks for the update. This is a useful hardening pass, but it spans several analyzer, provider, server, and cache paths, so the full review is continuing against the updated head rather than treating the earlier label as a verdict.

AlexsJones:
@Navneet072300 quick correction to my note from 08-27: you'd already split this into six focused PRs, so there's no separate full review of #1712 coming, sorry for the confusion. You mentioned closing this one in favour of the split, and it's still open (and now conflicts with main), so feel free to close it.

Where the split stands:

- #1734 and #1735 are merged, thanks!
- #1730 and #1731 are waiting on the regression tests asked for on 08-19. #1731 also needs a rebase.
- #1733 is waiting on tests and the `filepath.IsLocal` check before `xdg.CacheFile`.
- #1732 was approved but now conflicts with main, so it just needs a rebase.

Happy to re-review any of them as soon as they move.

PR #1717, opened 2026-07-28
fix(analyzer): skip non-Service HTTPRoute backends; serialize Gateway API scheme registration

### LINKED ISSUE #1: Add scenario for wrong selectorLabels in services (no endpoints)
## Description
When deploying services, one of the things which are faced very often is that endpoints are not available, caused by pods that are not able to start or incorrect labels on the service object.

## Possible solution
* Find services without endpoints
* If endpoint is missing, try to find out if pods with matching labels exist
* Look for problems on the pod
* If no pod matches labels, try to find pod with similar name
* Respond with problem solution

sunhyy:
## What this fixes

Two Gateway API analyzer bugs, both hit on clusters that use Envoy Gateway (or any non-Service HTTPRoute backend). Found on `v0.4.36`; the analyzer files are identical on `main`. The two fixes are independent and can be split into separate PRs if preferred.

### 1. HTTPRoute reports non-Service backends as missing Services

`HTTPRouteAnalyzer` looks up every `backendRef` as a core `Service` **without checking `Group`/`Kind`**, so a backend that references a non-Service resource (e.g. Envoy Gateway's `gateway.envoyproxy.io/Backend`, or a GRPC backend) is falsely reported:

```
HTTPRoute uses the Service 'ns/name' which does not exist.
```

Example backend that triggers it:

```yaml
backendRefs:
- group: gateway.envoyproxy.io
  kind: Backend
  name: my-backend
```

**Fix:** skip backendRefs whose `Group` is non-empty or whose `Kind != "Service"`, and nil-guard `*backend.Port`.

### 2. Concurrent Gateway API scheme registration crashes the process

`GatewayClass`, `Gateway` and `HTTPRoute` analyzers each call `gtwapi.AddToScheme(client.Scheme())` at the start of `Analyze`. Analyzers run concurrently (`max_concurrency`), so multiple goroutines write the shared `runtime.Scheme` maps at once:

```
fatal error: concurrent map writes
...
k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName(...)
sigs.k8s.io/gateway-api/apis/v1.addKnownTypes(...)
github.com/k8sgpt-ai/k8sgpt/pkg/analyzer.HTTPRouteAnalyzer.Analyze  pkg/analyzer/httproute.go:41
github.com/k8sgpt-ai/k8sgpt/pkg/analysis.(*Analysis).executeAnalyzer  pkg/analysis/analysis.go:428
```

Because this is a `fatal error` (not recoverable), the whole process exits (code 2) and restart-loops. It is most frequent on the instance that runs all three analyzers.

**Fix:** add `ensureGatewayScheme()`, which registers gateway-api types **at most once per scheme, under a mutex** — removing both the concurrent writes and the repeated writes. (An alternative would be to register the types when the ctrl client is built in `pkg/kubernetes`; this keeps the change local to the analyzers.)

## Testing

- Added `TestGWNonServiceBackendHTTPRouteAnalyzer` (an HTTPRoute whose backend is an Envoy Gateway `Backend`). It **fails without fix #1** — reproducing the false positive — and passes with it.
- `go build`, `go vet`, and `go test ./pkg/analyzer/` pass; existing Gateway/GatewayClass/HTTPRoute tests are unchanged.

### DIFF
--- a/pkg/analyzer/gateway.go
+++ b/pkg/analyzer/gateway.go
@@ -37,7 +37,7 @@ func (GatewayAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 	gtwList := &gtwapi.GatewayList{}
 	gc := &gtwapi.GatewayClass{}
 	client := a.Client.CtrlClient
-	err := gtwapi.AddToScheme(client.Scheme())
+	err := ensureGatewayScheme(client.Scheme())
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/analyzer/gateway_scheme.go
+++ b/pkg/analyzer/gateway_scheme.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package analyzer
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	gtwapi "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// The Gateway API analyzers (GatewayClass, Gateway, HTTPRoute) run concurrently
+// (see analysis.RunAnalysis with max_concurrency). Each used to call
+// gtwapi.AddToScheme(client.Scheme()) at the start of Analyze, mutating the shared
+// runtime.Scheme maps concurrently and triggering a fatal "concurrent map writes".
+//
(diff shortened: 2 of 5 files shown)


AlexsJones (changes requested):
Thanks for this, the non-Service backendRef fix is a real gap and the Envoy Gateway example makes it easy to verify. Two of the three changes here already have approved PRs in flight though: #1702 covers the nil `backend.Port` guard, and #1705 moves the gateway-api scheme registration to client construction (which also avoids the per-Analyze locking that `ensureGatewayScheme` needs). Since you kindly offered to split: could you narrow this PR to the Group/Kind skip in `httproute.go` plus its test, and rebase once those two land? That part is genuinely new and I'd like to get it in.

One mechanical note: the semantic-title check here only accepts the scope `deps`, so a plain `fix: ...` title (no scope) is needed to go green.

AlexsJones:
Hi @sunhyy, checking in on this one. Both of the overlapping PRs have landed since the review (#1702 on 08-12 and #1705 on 08-13), so the rebase is unblocked. Once it's narrowed to the Group/Kind skip in `httproute.go` plus its test, with a plain `fix: ...` title, it should be a quick review. Still keen to get the non-Service backendRef fix in if you have time.

PR #1722, opened 2026-08-07
fix: scan init containers for ConfigMap usage

### LINKED ISSUE #1720: [Bug] ConfigMap analyzer flags ConfigMaps used only by init containers as unused
## Description

The ConfigMap analyzer reports a ConfigMap as *"not used by any pods in the namespace"* even when it is actively referenced by an **init container** through `envFrom` or `env.valueFrom.configMapKeyRef`.

In `pkg/analyzer/configmap.go`, the usage scan only iterates over `pod.Spec.Containers`:

```go
// Check environment variables
for _, container := range pod.Spec.Containers {
    for _, env := range container.EnvFrom {
        if env.ConfigMapRef != nil {
            usedConfigMaps[env.ConfigMapRef.Name] = true
            ...
        }
    }
    for _, env := range container.Env {
        if env.ValueFrom != nil && env.ValueFrom.ConfigMapKeyRef != nil {
            usedConfigMaps[env.ValueFrom.ConfigMapKeyRef.Name] = true
            ...
        }
    }
}
```

`pod.Spec.InitContainers` is never scanned, so any ConfigMap referenced only from init containers is falsely flagged as unused. This is a common pattern for workloads that run setup/migration steps before the main containers start.

## Expected behavior

A ConfigMap referenced by an init container (via `envFrom` or `env.valueFrom.configMapKeyRef`) should be considered "used" and not reported.

## Actual behavior

```text
ConfigMap <name> is not used by any pods in the namespace
```

## Proposed fix

Scan `pod.Spec.InitContainers` with the same env/envFrom checks already applied to `pod.Spec.Containers`. A regression test covers both `envFrom` and `env.valueFrom.configMapKeyRef` cases.

## Environment

- k8sgpt version: main (reproduced with a fake client unit test)
- Kubernetes: any

junnhwan:
Closes #1720

## 📑 Description

The ConfigMap analyzer previously only walked `pod.Spec.Containers`. A ConfigMap referenced **exclusively** from an init container (via `envFrom` or `env.valueFrom.configMapKeyRef`) was therefore reported as unused.

This PR extends the usage scan to also iterate over `pod.Spec.InitContainers`, mirroring the existing checks for regular containers, and adds a regression test covering both env forms.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `gofmt` — clean
- `git diff --check` — clean

### DIFF
--- a/pkg/analyzer/configmap.go
+++ b/pkg/analyzer/configmap.go
@@ -60,6 +60,22 @@ func (ConfigMapAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 			}
 		}
 
+		// Check environment variables in init containers
+		for _, container := range pod.Spec.InitContainers {
+			for _, env := range container.EnvFrom {
+				if env.ConfigMapRef != nil {
+					usedConfigMaps[env.ConfigMapRef.Name] = true
+					configMapUsage[env.ConfigMapRef.Name] = append(configMapUsage[env.ConfigMapRef.Name], pod.Name)
+				}
+			}
+			for _, env := range container.Env {
+				if env.ValueFrom != nil && env.ValueFrom.ConfigMapKeyRef != nil {
+					usedConfigMaps[env.ValueFrom.ConfigMapKeyRef.Name] = true
+					configMapUsage[env.ValueFrom.ConfigMapKeyRef.Name] = append(configMapUsage[env.ValueFrom.ConfigMapKeyRef.Name], pod.Name)
+				}
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for the quick follow-through on #1720! Mirroring the existing `envFrom` and `env.valueFrom.configMapKeyRef` walk across `InitContainers` is exactly the right fix — same logic, same structure, zero ambiguity. The regression test covering both env forms is what you want here. CI green.

approve-recommend — ready for your final merge.

PR #1724, opened 2026-08-07
fix: skip empty results from custom analyzers with no findings

Q-Nimbus:
A custom analyzer with nothing to report returns `RunResponse.Result == nil`. `custom.Client.Run()` only populates its result when `Result != nil`, so it hands back the zero value — no `Name`, no `Error` — with a nil error, and `RunCustomAnalysis` appends that to `a.Results`.

### Why it matters

The appended entry is unusable to consumers, and it breaks **k8sgpt-operator**. `MapResults` derives `Result.metadata.name` from `result.Name`, so an empty name produces an object that fails CRD validation:

```
Result.core.k8sgpt.ai "" is invalid: [metadata.name: Required value, spec.error: Required value]
```

`processRawResults` does `return err` on the first failed create, so **one custom analyzer with no findings aborts the reconcile before the remaining results are written**. It ranges over a map, so the subset that survives differs every cycle, and other analyzers' Result CRs silently go stale — with nothing in the analyzer logs pointing at the cause. On our clusters this left Result CRs lagging from a few resourceVersions to tens of millions behind while every analyzer reported healthy counts.

### Why the analyzer cannot work around it

Both encodings of "no findings" are rejected:

- nil `Result` → empty `metadata.name`
- named `Result` with an empty `Error` list → still `spec.error: Required value`

So "I ran and found nothing" is currently unrepresentable without breaking the operator.

### The change

Skip the result when the analyzer returned nothing. This matches what the CLI already implies: an analyzer with no findings should contribute no entry. Analyzers that *do* report findings are unaffected, and a transport/analyzer error is still recorded in `a.Errors` as before.

### Tests

`pkg/analysis/analysis_test.go` now starts a real custom-analyzer gRPC server and asserts both directions:

- `TestRunCustomAnalysisNilResultProducesNoResult` — a nil `Result` yields no result **and** no error (no findings is not an error). Fails without this change with `Should be empty, but was [{EmptyAnalyzer  []  }]`.
- `TestRunCustomAnalysisPopulatedResultIsRecorded` — a populated `Result` is still recorded, with `Kind` still defaulting to the configured analyzer name.

`go test ./pkg/analysis/ ./pkg/custom/` passes.

### Note

A defensive guard in k8sgpt-operator (skipping a `resultSpec` with an empty `Name` in `MapResults`, or not aborting the whole loop on one bad result) would also be worth having, since it is the component that actually fails. Happy to open that separately if useful — fixing it here stops the empty entry being produced in the first place.

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -301,6 +301,27 @@ func (a *Analysis) RunCustomAnalysis() {
 			}
 
 			result, err := canClient.Run()
+			if err == nil && result.Name == "" && len(result.Error) == 0 {
+				// The analyzer returned RunResponse.Result == nil, i.e. "I ran and found nothing".
+				// custom.Client.Run() only populates its result when Result != nil, so what we hold
+				// here is the zero value: no Name, no Error.
+				//
+				// Appending it produces a result no consumer can use, and it is actively harmful to
+				// k8sgpt-operator: MapResults derives Result.metadata.name from result.Name, so an
+				// empty Name yields an object that fails CRD validation ("metadata.name: Required
+				// value, spec.error: Required value"). processRawResults returns on the first such
+				// failure, so ONE analyzer with no findings aborts the reconcile before the
+				// remaining results are written — and because it ranges over a map, the subset that
+				// survives differs every cycle.
+				//
+				// Skipping it is also what the CLI already implies: a custom analyzer with nothing
+				// to report should contribute nothing, not an empty entry.
+				if verbose {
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Nice catch on the operator reconcile abort — a nil `Result` producing an empty name that fails CRD validation and then short-circuits the entire reconcile loop makes this higher-severity than it looks from the analyzer side alone. The fix is minimal and correct: skip rather than append when the analyzer returns nothing, matching what CLI output already implies. Both test cases (nil result → no entry, populated → still recorded) cover the contract cleanly. CI green.

approve-recommend — ready for your final merge.

PR #1725, opened 2026-08-07
fix: skip failure report for jobs that completed successfully after retries

### LINKED ISSUE #1723: [Bug] Job analyzer flags successfully completed jobs that had retries as failed
## Description

The Job analyzer in `pkg/analyzer/job.go` reports a Job as failed whenever `job.Status.Failed > 0`, regardless of the Job's actual terminal state. Kubernetes Jobs have a retry mechanism driven by `spec.backoffLimit` (default 6): a pod may fail several times and then succeed on a later attempt. In that case the Job completes successfully (`status.conditions` contains `Complete=True`) but `status.failed` remains non-zero.

The analyzer never inspects `status.conditions`, so such healthy, successfully completed Jobs are reported as `"Job <name> has failed"` and produce false-positive alerts.

## Steps to reproduce

1. Create a Kubernetes Job with `spec.backoffLimit` set (e.g. 6) whose pod fails on the first attempt(s) and succeeds on a later attempt.
2. Wait until the Job reaches `Complete=True` (check with `kubectl get job <name> -o jsonpath='{.status.conditions}'`). Note that `status.failed` is non-zero.
3. Run `k8sgpt analyze` (or observe the k8sgpt-operator Result CRs).
4. k8sgpt reports the successfully completed Job as failed.

A unit test with a fake client reproduces it: a Job with `status.failed: 2`, `status.succeeded: 1` and a `Complete=True` condition currently produces one failure result.

## Expected behavior

A Job that has reached a successful terminal state should not be reported as failed, even when `status.failed` is non-zero (failed attempts that were retried within `backoffLimit` are normal).

## Actual behavior

```text
Job <name> has failed
```

fires for successfully completed Jobs that had retries.

## Root cause — analyzer code

In `pkg/analyzer/job.go`, the failure branch only checks the failure counter:

```go
if Job.Status.Failed > 0 {
    // reports "Job <name> has failed"
}
```

Kubernetes records the Job outcome in `status.conditions`:

- `Complete=True` (or `SuccessCriteriaMet=True`) — the Job succeeded;
- `Failed=True` — the Job actually failed (e.g. backoff limit exceeded).

`status.failed` is a cumulative counter of failed pod attempts, not a success/failure signal on its own.

## Suggested fix

Skip the failure report when the Job has reached a successful terminal state:

```go
if Job.Status.Failed > 0 && !jobHasSucceeded(Job) {
    // ...
}

// jobHasSucceeded reports whether the Job has reached a successful terminal state.
func jobHasSucceeded(job batchv1.Job) bool {
    for _, condition := range job.Status.Conditions {
        if condition.Status == corev1.ConditionTrue &&
            (condition.Type == batchv1.JobComplete || condition.Type == batchv1.JobSuccessCriteriaMet) {
            return true
        }
    }
    return false
}
```

A regression test covers both cases: a `Complete=True` Job with `status.failed > 0` is not reported, while a `Failed=True` Job is still reported.

## Environment

- k8sgpt version: main
- Kubernetes: any (reproduced with a fake client unit test)

junnhwan:
Closes #1723

## 📑 Description

The Job analyzer reported a Job as failed whenever `Status.Failed > 0`, without consulting the Job's terminal state in `status.conditions`. Kubernetes Jobs retry failed pods within `spec.backoffLimit` (default 6), so a Job that failed a few pods and then completed successfully still has `status.failed > 0` — and was wrongly reported as `"Job <name> has failed"`.

This PR guards the failure report with a new `jobHasSucceeded` helper that inspects `status.conditions` and returns true when either `Complete` or `SuccessCriteriaMet` (Kubernetes 1.31+) is `True`. A genuinely failed Job (`Failed=True`) is still reported.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `go test ./pkg/analyzer/ -count=1` — passes (including the two new regression tests)
- `git diff --check` — clean

Regression tests added:
- `TestJobAnalyzerSucceededAfterRetries` — `Complete=True` with `status.failed: 2` is not reported
- `TestJobAnalyzerFailedConditionOverridesRetries` — `Failed=True` is still reported

### DIFF
--- a/pkg/analyzer/job.go
+++ b/pkg/analyzer/job.go
@@ -19,6 +19,8 @@ import (
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/kubernetes"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/util"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
@@ -68,7 +70,11 @@ func (analyzer JobAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) 
 				},
 			})
 		}
-		if Job.Status.Failed > 0 {
+		// A Job that has reached a successful terminal state (Complete or
+		// SuccessCriteriaMet) should not be reported as failed even when
+		// Status.Failed is non-zero: failed attempts that were retried within
+		// backoffLimit are normal and the Job is healthy.
+		if Job.Status.Failed > 0 && !jobHasSucceeded(Job) {
 			doc := apiDoc.GetApiDocV2("status.failed")
 
 			failure := common.Failure{
@@ -114,4 +120,17 @@ func (analyzer JobAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) 
 	}
 
 	return a.Results, nil
+}
+
+// jobHasSucceeded reports whether the Job has reached a successful terminal
+// state. Kubernetes records a Complete condition when all pods succeed, and a
+// SuccessCriteriaMet condition when the Job's success policy is met.
+func jobHasSucceeded(job batchv1.Job) bool {
+	for _, condition := range job.Status.Conditions {
+		if condition.Status == corev1.ConditionTrue &&
+			(condition.Type == batchv1.JobComplete || condition.Type == batchv1.JobSuccessCriteriaMet) {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for the fast turnaround from the bug report to the fix. The `jobHasSucceeded` guard inspecting `status.conditions` rather than comparing counts is exactly right: Kubernetes keeps `status.failed` across retries, and the conditions are the real terminal state.

The two tests cover both ends well — retried-then-succeeded (no false positive) and genuinely failed (still reported). The `SuccessCriteriaMet` condition is a good forward-looking touch given 1.31+ Job success policies.

Build/lint/test workflows haven't run yet since this is your first contribution here. Once a maintainer approves the workflow run and CI comes back green, this is ready for merge.

AlexsJones (approved):
Nice follow-through on the job analyzer fix — the root cause is exactly right.

`jobHasSucceeded` is minimal and correct: it keys on `Complete` and `SuccessCriteriaMet` conditions, which is what Kubernetes uses to signal a healthy terminal state regardless of how many pod attempts it took. Adding `SuccessCriteriaMet` cleanly covers the 1.31+ success-policy path.

The two regression tests cover both directions (completed-with-retries not reported, genuinely-failed condition still reported). CI checks all green. Ready to merge.

PR #1728, opened 2026-08-10
fix: do not panic on an Ingress path with a resource backend

arpitjain099:
## 📑 Description

`IngressBackend` carries either a `Service` or a `Resource`, and exactly one of them is set. `Service` is a pointer, so on a path that uses a resource backend it is nil. The ingress analyzer does not check:

```go
for _, path := range rule.HTTP.Paths {
    _, err := a.Client.GetClient().CoreV1().Services(ing.Namespace).Get(a.Context, path.Backend.Service.Name, metav1.GetOptions{})
```

One resource-backed path anywhere in the namespace takes down the whole analyzer run:

```
panic: runtime error: invalid memory address or nil pointer dereference
analyzer.IngressAnalyzer.Analyze(...)
	pkg/analyzer/ingress.go:108
```

Resource backends are valid `networking.k8s.io/v1`, used for things like static assets served from an object-storage CRD, so this is reachable on a normal cluster rather than only on hand-crafted objects.

There is no service to look up for those paths, so the fix skips them and lets the rest of the ingress (class, TLS secrets, other paths) still be analyzed.

## ✅ Checks

## ℹ Additional Information

`TestIngressAnalyzer` gets a new "Resource backend instead of a service backend" case. It panics on main and passes with this change; `go test ./pkg/analyzer/` is green.


AlexsJones (approved):
Thanks for the clear write-up, the panic trace plus the "one resource-backed path takes down the whole run" framing is exactly what makes this easy to say yes to.

The guard is in the right place: all four `path.Backend.Service` dereferences in `pkg/analyzer/ingress.go` (the Get, the result text, and both mask fields) sit inside the loop you skip, and there is no `spec.defaultBackend` handling in this analyzer, so there is no second nil path left over. Skipping rather than reporting is the right call too, since a resource backend has nothing to validate against the Services API.

The test earns its keep by asserting the ingress-class finding still comes back, which is what proves the rest of the ingress is still analyzed instead of the whole object being skipped.

Approving for a maintainer merge. Tiny non-blocking nit: `resourceBackendAPIGroup` could live inside the test case as a local, since it exists only to give `APIGroup` something addressable, but a package-level var in a test file is harmless.

arpitjain099:
Thanks for the careful read, particularly checking that there is no `spec.defaultBackend` path left over. That was the thing I was least sure of when I scoped the guard.

Nit taken and pushed: `resourceBackendAPIGroup` is now a local at the top of `TestIngressAnalyzer` rather than a package-level var. You are right that it only exists to give `APIGroup` something addressable.

Separately, while checking whether the same class of problem showed up in the other analyzers, I found one in the HPA analyzer and opened #1729 for it. Different shape, not a panic: the `default` branch of the kind switch records "which is not an option" and then falls through to the `podInfo == nil` check, which records "which does not exist" for the same object. One unsupported kind produces two failures that contradict each other, and the second is wrong since nothing was ever looked up. Independent of this PR.

PR #1729, opened 2026-08-12
fix: do not also report an unsupported scaleTargetRef as missing

arpitjain099:
## 📑 Description

The HPA analyzer reports two contradictory failures for a single unsupported `scaleTargetRef.Kind`.

The kind switch has a `default` branch that records "which is not an option", but execution then falls into the `podInfo == nil` check below it, which records "which does not exist" as well. The second one is simply wrong: for an unsupported kind the analyzer never queried the API for anything, so it has no basis to say the target is missing.

Against an HPA with `kind: unsupported, name: foo`:

```
failure[0]: HorizontalPodAutoscaler uses unsupported as ScaleTargetRef which is not an option.
failure[1]: HorizontalPodAutoscaler uses unsupported/foo as ScaleTargetRef which does not exist.
```

Both go into `FailureDetails`, so the count reported by `AnalyzerErrorsMetric` is inflated too, and the LLM gets a self-contradicting pair of statements to explain.

The fix tracks whether the kind was one the analyzer actually looks up, and reports the missing target only in that case. `podInfo == nil` stays the outer condition, with the supported-kind test nested inside it, so the container checks in the `else` branch still run only when `podInfo` is non-nil. Guarding the combined condition instead (`podInfo == nil && supportedKind`) would send the unsupported-kind case into the `else` and nil-deref on `podInfo.GetPodSpec()`, which is worth mentioning since it is the obvious one-line version of this change.

The existing `TestHPAAnalyzerWithUnsuportedScaleTargetRef` passes either way, since it only checks that the "not an option" text is present somewhere and stops at the first match.

## ✅ Checks

## ℹ Additional Information

`TestHPAAnalyzerUnsupportedScaleTargetRefReportedOnce` is new and asserts exactly one failure with the right text. On main it fails with:

```
expected 1 failure, got 2: [HorizontalPodAutoscaler uses unsupported as ScaleTargetRef which is not an option. HorizontalPodAutoscaler uses unsupported/foo as ScaleTargetRef which does not exist.]
```

`go test ./pkg/analyzer/` is green with the change, `go vet` and `gofmt` clean.


AlexsJones (approved):
Thanks for the clear write-up and for calling out the nil-deref trap in the obvious one-liner. That context made the review fast.

The fix is correct: `supportedKind = false` in the `default` branch, then nesting the "does not exist" failure inside `if supportedKind`, means an unsupported kind records exactly one failure instead of two contradictory ones. The `else` branch stays safe since it only runs when `podInfo != nil`. The new test asserts exactly one failure with the right text, which is the right regression guard. CI green across all checks.

Recommending to the maintainer for merge.

AlexsJones (approved):
Nice focused regression test. The unsupported HPA target now produces its specific diagnostic without the misleading missing-target duplicate. Ready for your final look.

arpitjain099:
Thanks @AlexsJones. Nothing outstanding from my side.

The branch had gone behind main, so I have rebased it onto current main and force-pushed. No conflicts, and `go test ./pkg/analyzer/` is still green, so it should be clean for whenever the maintainer picks it up.

arpitjain099:
@AlexsJones I have brought the branch up to date with main. This has had your approval since 19 August, so it should be ready whenever you have a moment.

AlexsJones (approved):
Thanks for bringing this up to date. The current base-relative diff still suppresses only the unsupported-target duplicate, preserves the missing supported-target diagnostic, and keeps the container path behind the non-nil check. The focused one-failure regression and current build/lint checks are green. Ready for the maintainer's final look.

PR #1730, opened 2026-08-12
fix: implement graceful Shutdown RPC instead of panicking

Navneet072300:
## 📑 Description

Split out from #1712 per review feedback, since this is the one flagged as safe to merge on its own.

The `ServerConfigService.Shutdown` gRPC RPC was an unimplemented stub that called `panic("implement me")`. Since gRPC reflection is enabled on the server and this endpoint has no authentication, any client that can reach the port could crash a running `k8sgpt serve` process with a single call.

Wired the handler to a `ShutdownFunc` callback injected by the `server` package, so the listener lifecycle stays owned there rather than leaking into the handler. The RPC responds before the shutdown runs asynchronously, so it doesn't deadlock waiting on its own in-flight call closing the connection it's replying on.

## ✅ Checks

## ℹ Additional Information
No behavior changes beyond fixing the panic. `go build ./...`, `go vet ./...`, and the full test suite pass locally.

### DIFF
--- a/pkg/server/config/handler.go
+++ b/pkg/server/config/handler.go
@@ -4,13 +4,27 @@ import (
 	rpc "buf.build/gen/go/k8sgpt-ai/k8sgpt/grpc/go/schema/v1/schemav1grpc"
 	schemav1 "buf.build/gen/go/k8sgpt-ai/k8sgpt/protocolbuffers/go/schema/v1"
 	"context"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type Handler struct {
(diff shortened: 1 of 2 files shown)


AlexsJones (changes requested):
Thanks for replacing the panic with a graceful shutdown path. Please add a handler-level test that asserts the RPC returns , invokes the callback asynchronously, and returns  when no callback is configured.

AlexsJones (changes requested):
Thanks for replacing the panic with a graceful shutdown path. Please add a handler-level test that asserts the RPC returns the shutting-down response, invokes the callback asynchronously, and returns the Unimplemented status when no callback is configured.

PR #1731, opened 2026-08-12
fix: harden completion calls against empty responses and hangs

Navneet072300:
## 📑 Description

Two related robustness gaps in the AI backend clients:

- The OpenAI, Azure OpenAI, and Groq clients indexed `resp.Choices[0]` without checking the slice was non-empty. A content-filtered or safety-blocked response with zero choices panicked mid-analysis, losing every explanation already fetched in that run.
- None of the `http.Client` instances used by the OpenAI-compatible, custom REST, Ollama, and Anthropic clients set a `Timeout`, and the analysis context passed in has no deadline of its own. A hung AI endpoint blocked `analyze --explain` indefinitely with no way to abort short of `SIGKILL`.

Added a length check before indexing `Choices`, and a shared, generous (10 minute) default HTTP timeout — long enough not to interrupt slower local backends like Ollama — as a safety net.

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/pkg/ai/anthropic.go
+++ b/pkg/ai/anthropic.go
@@ -50,16 +50,16 @@ func (c *AnthropicClient) Configure(config IAIConfig) error {
 		opts = append(opts, option.WithAPIKey(token))
 	}
 
+	httpClient := &http.Client{Timeout: defaultHTTPTimeout}
 	proxyEndpoint := config.GetProxyEndpoint()
 	if proxyEndpoint != "" {
 		proxyURL, err := url.Parse(proxyEndpoint)
 		if err != nil {
 			return err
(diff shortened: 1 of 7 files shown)


AlexsJones (changes requested):
Thanks for hardening these client paths. Please add focused tests for the empty-choice error and for the configured timeout, especially a proxied client, so the new behavior is locked in across the affected SDK clients.

AlexsJones (changes requested):
Thanks for hardening these client paths. Please add focused tests for the empty-choice error and configured timeout, especially a proxied client, so the new behavior is locked in across the affected SDK clients.

PR #1732, opened 2026-08-12
fix: thread analyzer context instead of context.Background()

Navneet072300:
## 📑 Description

The Deployment, DaemonSet, and Mutating/ValidatingWebhook analyzers called the Kubernetes API with a hardcoded `context.Background()` instead of the caller-supplied `a.Context`, unlike every other analyzer in the package. This silently ignored any cancellation or timeout applied to the analysis context.

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/pkg/analyzer/daemonset.go
+++ b/pkg/analyzer/daemonset.go
@@ -1,7 +1,6 @@
 package analyzer
 
 import (
-	"context"
 	"fmt"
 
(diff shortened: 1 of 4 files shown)


AlexsJones (approved):
Reviewed the analyzer context propagation. The patch removes detached background contexts at each Kubernetes call site and consistently carries the caller cancellation/deadline. Ready for maintainer merge.

PR #1733, opened 2026-08-12
fix: prevent path traversal via cache key

Navneet072300:
## 📑 Description

`Exists`, `Load`, `Remove`, and `Store` in the file-based cache all built the cache file path by joining the k8sgpt cache directory with a caller-supplied key, without checking the result stayed inside that directory. A key such as `../../some-file` (e.g. passed to `k8sgpt cache purge`) could resolve outside the intended cache subdirectory.

Added a `cachePath()` helper that resolves the key and rejects any result that escapes the cache base directory, and used it in all four methods.

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/pkg/cache/file_based.go
+++ b/pkg/cache/file_based.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/adrg/xdg"
 	"github.com/k8sgpt-ai/k8sgpt/pkg/util"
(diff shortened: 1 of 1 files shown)


AlexsJones (changes requested):
Thanks for splitting this into a focused fix. Please add table-driven coverage for a normal key plus `../` and absolute-path escape attempts, including `Load`, `Store`, `Exists`, and `Remove`, so the boundary check stays protected. The semantic validation check is also red and will need resolving before this can land.

AlexsJones (changes requested):
Thanks for keeping this one narrow. The prefix check itself is right. Two things left:

- The table-driven tests from the earlier review are still missing. A normal key, `../` escapes and an absolute path through `Load`, `Store`, `Exists` and `Remove` would pin the check.
- `xdg.CacheFile` runs `os.MkdirAll` on the parent directory before `cachePath` checks the result (`pathutil.Create` in adrg/xdg v0.5.3). So `k8sgpt cache purge ../../evil/x` still creates `~/evil` before returning the error. Rejecting the key up front with `filepath.IsLocal(key)` avoids touching the filesystem, and the prefix check can stay as a backstop.

The branch is also behind main, so a rebase would get the full CI suite running on it.

PR #1735, opened 2026-08-12
fix: validate namespace exists before running core analyzers

Navneet072300:
## 📑 Description

Split out from #1712 per review feedback — this is the one flagged as needing discussion, so it's deliberately separate from the fixes that are safe to merge immediately.

`RunAnalysis`, the path used for all built-in analyzers, had no namespace validation: every analyzer's `List()` call against a nonexistent namespace just returns an empty list, so a typo'd `--namespace` silently reported "No problems detected" instead of an error.

Per review feedback, this only aborts on a confirmed `apierrors.IsNotFound(err)`. A service account scoped to a single namespace typically cannot `Get` the Namespace object itself, so a `Forbidden` (or any other) error falls through to the normal analysis path instead of blocking an analysis that would otherwise have succeeded.

Added coverage for both the NotFound-aborts and Forbidden-does-not-abort cases.

## ✅ Checks

## ℹ Additional Information
`go build ./...`, `go vet ./...`, and the full test suite pass locally. Happy to adjust the approach further based on discussion here.


AlexsJones (approved):
Reviewed the namespace validation. It only terminates on a confirmed NotFound, preserves restricted-service-account behavior, and covers both paths in tests. Ready for maintainer merge.

PR #1740, opened 2026-08-13
fix: report Deployment ProgressDeadlineExceeded when replica counts still match

### LINKED ISSUE #1739: [Bug] Deployment analyzer misses ProgressDeadlineExceeded when ReadyReplicas still match spec
## Expected Behavior

A Deployment whose rollout has failed (`status.conditions` type `Progressing`, status `False`, reason `ProgressDeadlineExceeded`) should be reported, even when `status.readyReplicas` still equals `spec.replicas`.

That is the state `kubectl describe deploy` shows for a stuck rollout: old Pods remain Ready, so the replica counts look healthy, but the Deployment has timed out progressing. See [Failed Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment).

## Current Behavior

`pkg/analyzer/deployment.go` only compares `spec.replicas` to `status.readyReplicas`. It never inspects `status.conditions`.

A common failed rollout therefore produces **no Deployment result**:

- `spec.replicas: 1` (or `maxUnavailable: 0`)
- new ReplicaSet never becomes ready (bad image, quota, probe, etc.)
- old Pod stays Ready → `status.readyReplicas == spec.replicas`
- `Progressing=False`, reason `ProgressDeadlineExceeded`

The replica check treats that as healthy. Existing tests only cover replica-count mismatch (and nil `spec.replicas`), so this path is untested.

## Steps to Reproduce

Reproduced with the fake-client tests already used in `pkg/analyzer/deployment_test.go` (no live cluster required):

1. Create a Deployment with `spec.replicas: 1`.
2. Set `status.readyReplicas: 1` and `status.availableReplicas: 1`.
3. Set conditions to the kubectl failed-rollout shape: `Available=True` (MinimumReplicasAvailable) and `Progressing=False` (ProgressDeadlineExceeded), e.g. message `ReplicaSet "example-6b7f8c9d5" has timed out progressing.`
4. Run the Deployment analyzer (`k8sgpt analyze --filter Deployment`, or the unit test harness).
5. Observe zero results for that Deployment.

Contrast: the same Deployment with `spec.replicas: 3` and `status.readyReplicas: 1` is reported, because only the replica counts are consulted.

## Environment

- **k8sgpt version:** main
- **Kubernetes version:** n/a (analyzer logic; reproduced with fake client). `ProgressDeadlineExceeded` is set by the Deployment controller after `spec.progressDeadlineSeconds` (default 600s).
- **OS/Platform:** macOS

## Additional Context

Current check:

```go
if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas != deployment.Status.ReadyReplicas {
    // report replica mismatch only
}
```

Suggested fix (out of scope: changing the replica-count logic; reporting `ReplicaFailure`, which the replica check often already covers): also report when `Progressing` is `False`; name the condition, reason, and message in the failure text.

junnhwan:
Closes #1739

## 📑 Description

The Deployment analyzer only compared `spec.replicas` to `status.readyReplicas`. A failed rollout can leave the old ReplicaSet Ready, so those counts still match while `Progressing=False` / `ProgressDeadlineExceeded` already records that the Deployment timed out.

This PR reports that condition by type and includes the reason and message in the failure text. Replica-count logic is unchanged. `Progressing=True` (healthy or in-progress) is not reported.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `go test ./pkg/analyzer/ -count=1 -run 'TestDeploymentAnalyzer'` — passes
- `gofmt` — clean
- `git diff --check` — clean

Regression tests added:
- `TestDeploymentAnalyzerProgressDeadlineExceeded` — `ReadyReplicas` matches spec but `Progressing=False` / `ProgressDeadlineExceeded` is reported with reason and message
- `TestDeploymentAnalyzerProgressingTrueNotReported` — completed rollout (`Progressing=True`) produces no result


AlexsJones (approved):
Nice catch on the stalled rollout case. The DeploymentProgressing false condition is surfaced independently of replica counts, with focused coverage for both false and true states. Ready for your final look.

PR #1741, opened 2026-08-13
fix: scan projected volume sources for ConfigMap usage

### LINKED ISSUE #1738: [Bug] ConfigMap analyzer reports projected volume ConfigMaps as unused
## Expected Behavior

A ConfigMap referenced from a Pod via a **projected volume source** (`spec.volumes[].projected.sources[].configMap`) should be treated as used, the same way a direct `spec.volumes[].configMap` reference is.

This includes the ConfigMap Kubernetes itself injects on every Pod: `kube-root-ca.crt` is mounted through the kubelet projected service-account volume, not through `volume.configMap`.

## Current Behavior

`pkg/analyzer/configmap.go` only records usage from `volume.ConfigMap`. Projected sources are ignored:

```go
for _, volume := range pod.Spec.Volumes {
    if volume.ConfigMap != nil {
        usedConfigMaps[volume.ConfigMap.Name] = true
        configMapUsage[volume.ConfigMap.Name] = append(configMapUsage[volume.ConfigMap.Name], pod.Name)
    }
}
```

So a ConfigMap that is only referenced under `projected.sources` is reported as:

```text
ConfigMap <name> is not used by any pods in the namespace
```

That is the same class of miss as #1720 (init-container env), just a different reference location.

On a default cluster this also false-positives `kube-root-ca.crt`. Kubernetes creates that ConfigMap in every namespace and the kubelet mounts it via a projected volume (`serviceAccountToken` + `configMap: kube-root-ca.crt` + `downwardAPI`). See [Projected Volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes/) and the default [service account token projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#projected-volume).

Existing tests only cover `envFrom` / `env.valueFrom.configMapKeyRef` (including init containers after #1722). There is no projected-volume case.

## Steps to Reproduce

Reproduced with the fake-client tests already used in `pkg/analyzer/configmap_initcontainer_test.go` (no live cluster required):

1. Create ConfigMap `projected-cm` with some data.
2. Create a Pod whose only reference is:

```yaml
volumes:
- name: cfg
  projected:
    sources:
    - configMap:
        name: projected-cm
```

3. Run the ConfigMap analyzer (`k8sgpt analyze --filter ConfigMap`, or the unit test harness).
4. Observe: `ConfigMap projected-cm is not used by any pods in the namespace`.

Same result with the kubelet default shape: ConfigMap `kube-root-ca.crt` plus a projected volume containing `serviceAccountToken`, `configMap.name=kube-root-ca.crt`, and `downwardAPI`.

Contrast: the same ConfigMap referenced via `volumes[].configMap` is not reported.

## Environment

- **k8sgpt version:** main
- **Kubernetes version:** n/a (analyzer logic; reproduced with fake client). The `kube-root-ca.crt` projected mount is default since Kubernetes 1.22.
- **OS/Platform:** macOS

## Additional Context

Suggested fix (out of scope: scanning Deployments/Jobs that have no running Pods, Secret projected sources, skipping `kube-root-ca.crt` by name):

When walking `pod.Spec.Volumes`, also mark `volume.Projected.Sources[].ConfigMap.Name` as used.

junnhwan:
Closes #1738

## 📑 Description

The ConfigMap analyzer only counted `volumes[].configMap`. A ConfigMap referenced solely via `projected.sources[].configMap` was reported as unused. That includes `kube-root-ca.crt`, which the kubelet mounts on every Pod through the default projected service-account volume.

This PR walks `volume.Projected.Sources[].ConfigMap` and leaves the existing direct `volume.ConfigMap` accounting intact.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `go test ./pkg/analyzer/ -count=1 -run 'TestConfigMapAnalyzer'` — passes
- `gofmt` — clean
- `git diff --check` — clean

Regression tests added:
- `TestConfigMapAnalyzer_ProjectedVolumeUsage` — ConfigMap used only via a projected source is not reported unused
- kubelet-style `kube-root-ca.crt` projected volume is treated as used
- an unused ConfigMap is still reported


AlexsJones (approved):
Nice focused fix for projected volumes. ConfigMap projections are now counted as usage and the regression coverage includes the service-account projection shape. Ready for your final look.

PR #1743, opened 2026-08-19
fix: ignore in-progress Deployment replica mismatches

### LINKED ISSUE #1742: [Bug] Deployment analyzer reports in-progress rollouts as failures
## Expected Behavior

A Deployment that is **actively progressing** (rollout or scaling in progress) should not be reported as a problem. `status.readyReplicas` trailing `spec.replicas` is a normal, healthy intermediate state while the Deployment controller is rolling out a new ReplicaSet or scaling up — image pulls and readiness probes take time, and `maxSurge`/`maxUnavailable` deliberately allow temporarily fewer ready pods.

## Current Behavior

`pkg/analyzer/deployment.go` reports a failure whenever `spec.replicas != status.readyReplicas`, without inspecting the Deployment's progress state. During a normal rollout (e.g. `spec.replicas: 3`, `status.replicas: 3`, `status.readyReplicas: 2`), running `k8sgpt analyze --filter Deployment` emits:

```
Deployment default/web has 3 replicas but 2 are available with status running
```

This is a false positive: the third pod is still pulling the image / waiting on its readiness probe, and the Deployment is healthy.

The check also does not consider `status.observedGeneration` vs `metadata.generation`. When the controller has not yet reconciled the latest spec (`observedGeneration < generation`), status is stale relative to spec, and the replica comparison is misleading. No analyzer in `pkg/analyzer/` currently consults `observedGeneration`.

## Steps to Reproduce

Reproduces purely at the status level with the fake client already used in `pkg/analyzer/deployment_test.go` (no live cluster required):

1. Create a Deployment with `spec.replicas: 3` and a pod template.
2. Set `status.replicas: 3`, `status.readyReplicas: 2`, `status.availableReplicas: 2` — the mid-rollout shape.
3. Run the Deployment analyzer (`k8sgpt analyze --filter Deployment`, or the unit test harness).
4. Observe the failure text above, even though nothing is wrong.

To reproduce on a live cluster: deploy a Deployment with a readiness probe, update its image, and run `k8sgpt analyze` during the rollout window (seconds to minutes depending on image size and probe time).

## Environment

- **k8sgpt version:** main
- **Kubernetes version:** n/a (analyzer logic; reproduced with fake client). Rollout semantics are identical across versions.
- **OS/Platform:** n/a

## Additional Context

Current check:

```go
if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas != deployment.Status.ReadyReplicas {
    // report replica mismatch only
}
```

Suggested direction for the fix (open to maintainer preference; deliberately does not overlap with the `Progressing=False` / `ProgressDeadlineExceeded` reporting from #1739 / #1740, which covers the *stuck* rollout case):

1. **`ObservedGeneration` guard**: skip the replica comparison while `status.observedGeneration < metadata.generation` (controller hasn't caught up to the latest spec yet).
2. **Healthy-progress check**: when `status.conditions` show `Progressing=True` (with `Available=True`), treat the replica mismatch as in-progress and do not report; only report when the rollout is genuinely stuck or replicas are unavailable.
3. Add regression tests for: healthy rollout-in-progress, scaling-in-progress, controller-not-yet-synced (`observedGeneration < generation`), and a genuinely stuck rollout (to keep the existing #1739 behavior).

Scope: single file `pkg/analyzer/deployment.go` + `deployment_test.go`.

junnhwan:
Closes #1742

## 📑 Description

The Deployment analyzer reports replica mismatches during healthy rollouts because it compares `spec.replicas` with `status.readyReplicas` without checking status freshness or progress state. This change skips that mismatch while the controller has not observed the latest generation or the Deployment reports both `Available=True` and `Progressing=True`. Existing `Progressing=False` / `ProgressDeadlineExceeded` failures remain reported.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `go test ./pkg/analyzer/ -count=1 -run 'TestDeploymentAnalyzer'` — passes
- `gofmt` — clean
- `git diff --check` — clean

Regression tests added:
- `TestDeploymentAnalyzerInProgressRolloutsNotReported`
- Existing `TestDeploymentAnalyzerProgressDeadlineExceeded` remains in place

### DIFF
--- a/pkg/analyzer/deployment.go
+++ b/pkg/analyzer/deployment.go
@@ -56,7 +56,7 @@ func (d DeploymentAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) 
 
 	for _, deployment := range deployments.Items {
 		var failures []common.Failure
-		if deployment.Spec.Replicas != nil && *deployment.Spec.Replicas != deployment.Status.ReadyReplicas {
+		if shouldReportReplicaMismatch(deployment) {
 			if deployment.Status.Replicas > *deployment.Spec.Replicas {
 				doc := apiDoc.GetApiDocV2("spec.replicas")
 
@@ -140,4 +140,36 @@ func (d DeploymentAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) 
 	}
 
 	return a.Results, nil
+}
+
+// shouldReportReplicaMismatch ignores stale status and healthy in-progress rollouts.
+func shouldReportReplicaMismatch(deployment appsv1.Deployment) bool {
+	if deployment.Spec.Replicas == nil || *deployment.Spec.Replicas == deployment.Status.ReadyReplicas {
+		return false
+	}
+
+	if deployment.Status.ObservedGeneration < deployment.Generation {
+		return false
+	}
+
+	if hasHealthyProgress(deployment.Status.Conditions) {
+		return false
+	}
+
+	return true
+}
+
+func hasHealthyProgress(conditions []appsv1.DeploymentCondition) bool {
+	progressing := false
+	available := false
+	for _, condition := range conditions {
+		switch condition.Type {
+		case appsv1.DeploymentProgressing:
+			progressing = condition.Status == corev1.ConditionTrue
+		case appsv1.DeploymentAvailable:
+			available = condition.Status == corev1.ConditionTrue
+		}
+	}
+
+	return progressing && available
 }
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for this, and specifically for gating on `ObservedGeneration` as well as the conditions. That is the half people usually miss, and it is what stops a freshly-edited Deployment from being judged on status the controller has not written yet.

Requiring both `Available=True` and `Progressing=True` is the right conjunction. Progressing alone would silence a rollout that has never become available, which is exactly the case worth reporting.

Worth stating explicitly since it is the deliberate trade-off here: a genuinely stuck rollout stays quiet until `progressDeadlineSeconds` elapses and the condition flips to `Progressing=False` / `ProgressDeadlineExceeded`. That is 10 minutes on the default, and it is the correct call (a false positive on every deploy is worse than a delayed true positive), but it is a behaviour change worth a line in the release notes.

Table-driven tests cover all three suppression paths and the ProgressDeadlineExceeded case still reports. Full CI green including codecov patch. Recommending merge.

PR #1747, opened 2026-08-21
fix: scope NetworkPolicy pod lookup to policy namespace

### LINKED ISSUE #1746: [Bug] NetworkPolicy analyzer matches same-label Pods from other namespaces
## Expected Behavior

When analyzing NetworkPolicies across all namespaces, a policy's `spec.podSelector` should be evaluated only against Pods in the policy's own namespace. A Pod with matching labels in another namespace must not count as a selected Pod.

## Current Behavior

`NetworkPolicyAnalyzer` passes the analysis namespace to the Pod lookup instead of the NetworkPolicy namespace. When analysis runs without `--namespace`, the analysis namespace is empty and the client lists Pods across namespaces. As a result, a policy in one namespace can be treated as applied because a same-label Pod exists in another namespace, suppressing the diagnostic that the policy selects no Pods in its own namespace.

## Steps to Reproduce

1. Create a NetworkPolicy in `team-a` with `spec.podSelector.matchLabels: {app: worker}` and no matching Pod in `team-a`.
2. Create a Pod with label `app: worker` only in `team-b`.
3. Run the NetworkPolicy analyzer in cluster-wide mode by omitting `--namespace` (or call `NetworkPolicyAnalyzer` with `common.Analyzer.Namespace == ""`).
4. Observe that the `team-a/worker-policy` result is absent, even though the policy does not select any Pod in `team-a`.

The same behavior is reproducible with the repository's fake client: a `team-a` NetworkPolicy and a same-label `team-b` Pod produce zero analyzer results when `Namespace` is empty.

## Environment

- **k8sgpt version:** main
- **Kubernetes version:** networking.k8s.io/v1 NetworkPolicy semantics; not version-specific

## Additional Context

The analyzer currently lists policies using the analysis namespace and then calls the Pod helper with `a.Namespace`, while NetworkPolicy `podSelector` is namespace-scoped by Kubernetes semantics. The CLI defaults `--namespace` to an empty value, so this affects the normal cluster-wide analysis path.

Kubernetes API reference: https://kubernetes.io/docs/reference/kubernetes-api/networking/network-policy-v1/

A temporary fake-client regression test failed deterministically on the current checkout. The focused scope of a possible fix is to evaluate the selector in `policy.Namespace` and add the regression test. Support for `matchExpressions` is a separate behavior and is intentionally out of scope for this issue.

junnhwan:
Closes #1746

## 📑 Description

When NetworkPolicy analysis runs cluster-wide, the analyzer can match a policy in one namespace against a same-label Pod in another namespace. This suppresses the diagnostic that the policy selects no Pods in its own namespace because the Pod lookup uses the analysis namespace instead of the policy namespace.

This change scopes the Pod lookup to `policy.Namespace` and adds a fake-client regression test covering a `team-a` policy with only a same-label `team-b` Pod. Support for `matchExpressions` is intentionally outside this change.

## ✅ Checks

## ℹ Additional Information

Verified locally:

- `go test ./pkg/analyzer/ -count=1 -run '^TestNetpol'` — passes
- `gofmt -w pkg/analyzer/netpol.go pkg/analyzer/netpol_test.go` — clean
- `git diff --check` — clean

Regression tests added:

- `TestNetpolNoPodsUsesPolicyNamespace` — verifies that a same-label Pod in another namespace does not satisfy a NetworkPolicy's selector.

### DIFF
--- a/pkg/analyzer/netpol.go
+++ b/pkg/analyzer/netpol.go
@@ -69,7 +69,7 @@ func (NetworkPolicyAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error)
 			})
 		} else {
 			// Check if policy is not applied to any pods
-			podList, err := util.GetPodListByLabels(a.Client.GetClient(), a.Namespace, policy.Spec.PodSelector.MatchLabels)
+			podList, err := util.GetPodListByLabels(a.Client.GetClient(), policy.Namespace, policy.Spec.PodSelector.MatchLabels)
 			if err != nil {
 				return nil, err
 			}
--- a/pkg/analyzer/netpol_test.go
+++ b/pkg/analyzer/netpol_test.go
@@ -134,6 +134,59 @@ func TestNetpolWithPod(t *testing.T) {
 	assert.Equal(t, len(results), 0)
 }
 
+func TestNetpolNoPodsUsesPolicyNamespace(t *testing.T) {
+	clientset := fake.NewSimpleClientset(
+		&networkingv1.NetworkPolicy{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "worker-policy",
+				Namespace: "team-a",
+			},
+			Spec: networkingv1.NetworkPolicySpec{
+				PodSelector: metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"app": "worker",
+					},
+				},
+			},
+		},
+		&v1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "worker",
+				Namespace: "team-b",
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Nice catch, this is the same cluster-wide namespace slip as the Service analyzer one. Looking Pods up in `policy.Namespace` is right: the policies are listed from `a.Namespace`, which is empty in a cluster-wide run, so a same-label Pod in another namespace was hiding the "not applied to any pods" finding. `TestNetpolNoPodsUsesPolicyNamespace` fails without the change, and the full CI run on this head is green. Main still has the old lookup, and the branch merges cleanly. Good to merge.

Issue #1074, opened 2024-04-07
Getting error "AI provider openai not specified in configuration. Please run k8sgpt auth" while using k8sgpt cli with Azureopenai as backend


venkatest:
### Checklist

- [X] I've searched for similar issues and couldn't find anything matching
- [X] I've included steps to reproduce the behavior

### Affected Components

- [X] K8sGPT (CLI)
- [ ] K8sGPT Operator

### K8sGPT Version

v0.3.29

### Kubernetes Version

v1.29.1

### Host OS and its Version

Windows

### Steps to reproduce

![image](https://github.com/k8sgpt-ai/k8sgpt-operator/assets/51379199/ddaa8df3-4922-4342-bbe8-381ecb77b4eb)

k8sgpt auth default -p azureopenai

k8sgpt auth add --backend azureopenai --baseurl https://xxxxxx.openai.azure.com --engine gpt3-5 --model gpt-35-turbo

![image](https://github.com/k8sgpt-ai/k8sgpt-operator/assets/51379199/a1386e33-fbd5-44c8-ac55-030a2366ea1e)

![image](https://github.com/k8sgpt-ai/k8sgpt-operator/assets/51379199/389050d0-ec4c-4cf4-96ff-49a057b902e9)

k8sgpt analyze --explain --filter=Pod

![image](https://github.com/k8sgpt-ai/k8sgpt-operator/assets/51379199/1c98355a-7eba-48d5-bb2e-5d0f97b4196d)

### Expected behaviour

K8sgpt should give analyze results

### Actual behaviour

Error: AI provider openai not specified in configuration. Please run k8sgpt auth

### Additional Information

I checked this issue : https://github.com/k8sgpt-ai/k8sgpt-operator/issues/98 but didnt help

strowi:
Having a similar issue with v0.3.40 + local ollama:

```bash
~> ollama serve &
~ » k8sgpt auth add --backend localai --model mistral --baseurl http://localhost:11434/v1
localai added to the AI backend provider list

~ » k8sgpt analyze --explain --filter=Service
Error: AI provider openai not specified in configuration. Please run k8sgpt auth
```

raghu25:
try doing to switch to the desired backend

 k8sgpt auth default -p<backend ai>

jlcrow:
[...]

PR #1749, opened 2026-08-21
fix: honour K8SGPT_BACKEND on every serve start

michal-ks:
Relates to #1074

## 📑 Description

`serve` reads `K8SGPT_BACKEND` only when the config holds no providers yet, and
it persists the provider it resolves on first start. Where that config outlives
the process, as it does under the operator which points `XDG_CONFIG_HOME` at a
volume that survives container restarts, every later start finds a populated
config, skips the env block, and leaves `backend` at the `--backend` default of
`openai`. The lookup then fails against a config that never contained `openai`:

```
Error: AI provider openai not specified in configuration. Please run k8sgpt auth
```

So a single container restart breaks the deployment permanently, and deleting
the pod is the only recovery, since the operator exposes no way to pass
`--backend`. It also only reproduces on the *second* start, which is why a clean
install looks fine. Matches the operator report in #1074 (comment of
2026-04-12).

Verified on `main` (05247a8) with `XDG_CONFIG_HOME` set and
`K8SGPT_BACKEND=google`:

```
run 1  starts, writes provider "google" to the config
run 2  Error: AI provider openai not specified in configuration   (exit 1)
```

With this change run 2 starts normally. `--backend openai` still overrides the
environment, and with no env set the error is byte-identical to before, so
precedence is unchanged.

## ✅ Checks

## ℹ Additional Information

`TestResolveBackendPrefersEnvOverFlagDefault` fails on `main` with `got "openai"`
and passes with the fix. Two further tests pin the explicit-flag and unset-env
cases so the precedence cannot regress quietly.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -59,6 +59,9 @@ var ServeCmd = &cobra.Command{
 			color.Red("Error: %v", err)
 			os.Exit(1)
 		}
+		// K8SGPT_BACKEND must be honoured on every start, not only the first.
+		backend = resolveBackend(cmd.Flags().Changed("backend"), backend)
+
 		var aiProvider *ai.AIProvider
 		if len(configAI.Providers) == 0 {
 			// we validate and set temperature for our backend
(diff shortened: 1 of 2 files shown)


AlexsJones (commented):
Thanks for tracking this down, the "only breaks on the second start" part is a nasty one. The fix is right: `providerFromEnv` is the only place that reads `K8SGPT_BACKEND`, and it only runs while the config has no providers, so every later start falls back to the `openai` flag default. Resolving the backend before that block, with an explicit `--backend` still winning, fixes it without changing the first-start path, and the three `resolveBackend` tests cover the cases that matter. The code looks ready to me. The test, lint and build runs on this head are still waiting for maintainer approval, so this needs those green before it merges.

PR #1750, opened 2026-08-23
fix: report ReplicaFailure regardless of replica count

### LINKED ISSUE #1748: [Bug] ReplicaSet analyzer ignores ReplicaFailure when replicas already exist
## Expected Behavior

The ReplicaSet analyzer should report a ReplicaSet when its status contains a `ReplicaFailure=True` condition with reason `FailedCreate`, regardless of how many replicas have already been observed. A partially fulfilled ReplicaSet can still be failing to create additional Pods.

## Current Behavior

`ReplicaSetAnalyzer` only inspects `status.conditions` inside `if rs.Status.Replicas == 0`. When a ReplicaSet has some existing Pods but subsequent Pod creations fail, Kubernetes can expose `status.replicas > 0` together with `ReplicaFailure=True`; the analyzer then returns no ReplicaSet result and suppresses the failure message.

## Steps to Reproduce

1. Create a ReplicaSet with `spec.replicas: 1` and a valid Pod template.
2. After its first Pod is running, scale the ReplicaSet to `3` while a ResourceQuota or admission policy allows the existing Pod but rejects additional Pod creations.
3. Observe a status shape like:

   ```yaml
   status:
     replicas: 1
     conditions:
     - type: ReplicaFailure
       status: "True"
       reason: FailedCreate
       message: quota exceeded
   ```

4. Run the ReplicaSet analyzer (`k8sgpt analyze --filter ReplicaSet`).
5. Observe that no result is produced for the ReplicaSet, even though its status reports a failed Pod creation.

The same behavior is reproducible with the repository's fake client: a ReplicaSet with desired replicas `3`, `status.replicas: 1`, and a `ReplicaFailure=True/FailedCreate` condition produces zero analyzer results on the current checkout.

## Environment

- **k8sgpt version:** main
- **Kubernetes version:** `apps/v1` ReplicaSet status semantics; not version-specific

## Additional Context

The current gate is in `pkg/analyzer/rs.go`: the analyzer only walks `status.conditions` when `status.replicas == 0`. Kubernetes documents `status.replicas` as the current number of non-terminating Pods, while `status.conditions` represents the latest observations of the ReplicaSet state. The `ReplicaFailure` condition is added when a Pod fails to be created or deleted; it is not defined as an empty-ReplicaSet-only condition.

Kubernetes API reference: https://kubernetes.io/docs/reference/kubernetes-api/apps/replica-set-v1/

A temporary fake-client regression test failed deterministically on `origin/main` (`05247a8`). Removing only the `status.replicas == 0` gate makes the new regression pass along with the existing ReplicaSet tests. This issue focuses on the existing `FailedCreate` path; other ReplicaFailure reasons and event aggregation are out of scope.

No matching open Issue or PR was found for `ReplicaFailure`, `rs.go`, or this behavior.

junnhwan:
## 📑 Description

The ReplicaSet analyzer only inspected `status.conditions` inside `if rs.Status.Replicas == 0`, so a partially fulfilled ReplicaSet that was failing to create additional Pods (`status.replicas > 0` together with `ReplicaFailure=True`/`FailedCreate`) suppressed the failure and produced no result.

`ReplicaFailure` is reported by the ReplicaSet controller when a Pod fails to be created, independent of how many Pods already exist. This change evaluates `ReplicaFailure=True`/`FailedCreate` independently of the replica count. Scope is kept to the `FailedCreate` reason as a focused first step.

Closes #1748

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `gofmt` — clean
- `git diff --check` — clean

Regression tests added:
- `TestReplicaSetAnalyzerReplicaFailureWithExistingReplicas` covers a ReplicaSet with one existing replica plus a failed scale-up (`ReplicaFailure=True`/`FailedCreate`), and a healthy partially available control case.

### DIFF
--- a/pkg/analyzer/rs.go
+++ b/pkg/analyzer/rs.go
@@ -42,20 +42,19 @@ func (ReplicaSetAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 	for _, rs := range list.Items {
 		var failures []common.Failure
 
-		// Check for empty rs
-		if rs.Status.Replicas == 0 {
-
-			// Check through container status to check for crashes
-			for _, rsStatus := range rs.Status.Conditions {
-				if rsStatus.Type == "ReplicaFailure" && rsStatus.Reason == "FailedCreate" {
-					failures = append(failures, common.Failure{
-						Text:      rsStatus.Message,
-						Sensitive: []common.Sensitive{},
-					})
-
-				}
+		// ReplicaFailure is reported by the ReplicaSet controller when a Pod
+		// fails to be created, independent of how many Pods already exist.
+		// Evaluate it regardless of the current replica count so a partially
+		// fulfilled ReplicaSet can still surface a failed scale-up.
+		for _, rsStatus := range rs.Status.Conditions {
+			if rsStatus.Type == "ReplicaFailure" && rsStatus.Reason == "FailedCreate" {
+				failures = append(failures, common.Failure{
+					Text:      rsStatus.Message,
+					Sensitive: []common.Sensitive{},
+				})
 			}
 		}
+
 		if len(failures) > 0 {
 			preAnalysis[fmt.Sprintf("%s/%s", rs.Namespace, rs.Name)] = common.PreAnalysis{
 				ReplicaSet:     rs,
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for keeping this focused. I checked that `ReplicaFailure` is now evaluated independently of replica count and that the regression covers both the partially fulfilled failed-create case and the healthy control. CI is green; ready for your final look.

PR #1753, opened 2026-08-24
fix: scope Service events to their object identity

### LINKED ISSUE #1752: [Bug] Service analyzer attributes warning Events to the wrong object
## Expected Behavior

The Service analyzer should report a Warning Event only when the Event's `involvedObject` identifies the Service being analyzed. The identity must include the object kind, namespace, and name.

A healthy `team-a/api` Service must not inherit a Warning Event for a same-name Pod, or for a same-name Service in `team-b`, during cluster-wide analysis.

## Current Behavior

`ServiceAnalyzer` lists Events with `involvedObject.name=<endpoint name>` but uses the analysis namespace (`a.Namespace`) and does not constrain the involved object kind. When analysis runs cluster-wide, `a.Namespace` is empty, so Events from every namespace can be considered. A same-name Pod or Service can therefore add its Warning message to an unrelated Service result.

This creates false positives and can hide the actual source of a warning: the analyzer presents an Event for another Kubernetes object as if it came from the Service.

## Steps to Reproduce

1. Create a Service named `api` in `default` with a ready Endpoint, so the Service itself has no endpoint failure.
2. Create an unrelated Pod named `api` in the same namespace and make it emit a Warning Event, for example by using an invalid image. The Event should have:

   ```yaml
   involvedObject:
     kind: Pod
     namespace: default
     name: api
   type: Warning
   message: unrelated Pod warning
   ```

3. Run the Service analyzer with `k8sgpt analyze --filter Service`.
4. Observe that the analyzer reports the Service with the Pod's Warning Event, even though the Service has a ready Endpoint.

The same behavior occurs during cluster-wide analysis when `team-a/api` is healthy but a Warning Event for `team-b/api` exists.

The repository's fake client reproduces both cases deterministically on `origin/main` (`05247a8`):

- `TestServiceAnalyzerIgnoresEventsForOtherKinds` returns `Service default/api has event unrelated Pod warning`.
- `TestServiceAnalyzerScopesEventsToEndpointNamespace` returns `Service team-a/api has event unrelated Service warning`.

## Environment

- **k8sgpt version:** main (`05247a8`)
- **Kubernetes version:** core/v1 Event and namespace-scoped Event-list semantics; not version-specific

## Additional Context

The relevant code is in `pkg/analyzer/service.go:117-130`. Kubernetes defines `Event.involvedObject` as the object the Event is about, and Event field selectors support `involvedObject.kind`, `involvedObject.namespace`, and `involvedObject.name`:

- https://kubernetes.io/docs/reference/kubernetes-api/core/event-v1/
- https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

A focused fix can scope the Event list to `ep.Namespace` and require `event.InvolvedObject.Kind == "Service"` before adding a failure. In an isolated temporary worktree, that two-part change made both new regression cases pass, the existing Service analyzer tests pass, `go test ./... -count=1`, `go vet ./pkg/analyzer/`, and `git diff --check` pass.

This Issue is limited to Event-to-Service identity. It does not change the shared `util.FetchLatestEvent` helper or the Pod, Job, PVC, and StatefulSet analyzers; those callers can be audited separately.

No matching open Issue or PR was found for `Service analyzer`, `involvedObject`, `FetchLatestEvent`, `service.go`, or this behavior. Closed Issue #1671 / PR #1672 concern a different Service analyzer panic caused by a nil Endpoint `TargetRef`.

junnhwan:
Closes #1752

## 📑 Description

The Service analyzer lists Events using the analysis namespace and only matches `involvedObject.name`. During cluster-wide analysis, this can attach a Warning Event from another namespace or another Kubernetes kind to a Service with the same name.

This PR scopes the Event list to the Endpoint namespace and requires `involvedObject.kind == "Service"` before adding the failure. The existing name selector remains in place, so the Service Event identity is matched without changing the shared Event helper or other analyzers.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `gofmt` — clean
- `git diff --check` — clean

Regression tests added:
- `TestServiceAnalyzerIgnoresEventsForOtherKinds` — a same-name Pod Warning Event does not create a Service result
- `TestServiceAnalyzerScopesEventsToEndpointNamespace` — a same-name Service Warning Event in another namespace does not create a result during cluster-wide analysis

### DIFF
--- a/pkg/analyzer/service.go
+++ b/pkg/analyzer/service.go
@@ -114,7 +114,7 @@ func (ServiceAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 			}
 		}
 		// fetch event
-		events, err := a.Client.GetClient().CoreV1().Events(a.Namespace).List(a.Context,
+		events, err := a.Client.GetClient().CoreV1().Events(ep.Namespace).List(a.Context,
 			metav1.ListOptions{
 				FieldSelector: "involvedObject.name=" + ep.Name,
 			})
@@ -123,7 +123,7 @@ func (ServiceAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 			return nil, err
 		}
 		for _, event := range events.Items {
-			if event.Type != "Normal" {
+			if event.Type != "Normal" && event.InvolvedObject.Kind == kind {
 				failures = append(failures, common.Failure{
 					Text: fmt.Sprintf("Service %s/%s has event %s", ep.Namespace, ep.Name, event.Message),
 				})
--- a/pkg/analyzer/service_test.go
+++ b/pkg/analyzer/service_test.go
@@ -402,3 +402,76 @@ func TestServiceAnalyzerLabelSelectorFiltering(t *testing.T) {
 	require.Equal(t, 1, len(results))
 	require.Equal(t, "default/Endpoint1", results[0].Name)
 }
+
+func TestServiceAnalyzerIgnoresEventsForOtherKinds(t *testing.T) {
+	clientSet := fake.NewSimpleClientset(
+		&v1.Endpoints{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "api",
+				Namespace: "default",
+			},
+			Subsets: []v1.EndpointSubset{{
+				Addresses: []v1.EndpointAddress{{
+					IP: "10.0.0.1",
+				}},
+			}},
+		},
+		&v1.Event{
+			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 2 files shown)


git-jxj:
Thank you for the original focused fix and regression cases. Since #1783 has now merged the kind/name field selector and this branch is 21 commits behind `main`, I opened #1795 from the current codebase to carry forward the remaining namespace-scoping fix (`Events(ep.Namespace)`) without reverting the selector already on `main`. Maintainers can treat #1795 as the current-main replacement for the remaining part of this PR.

PR #1755, opened 2026-08-25
fix: resolve ClusterRole references in RoleBindings

### LINKED ISSUE #1754: [Bug] Security analyzer resolves ClusterRole RoleBindings as namespaced Roles
## Expected Behavior

The Security analyzer should resolve a `RoleBinding` according to
`roleRef.kind`:

- `Role` should be read from `Roles(<roleBinding namespace>)`.
- `ClusterRole` should be read from the cluster-scoped `ClusterRoles` resource.

It should then inspect the referenced object's rules for wildcard verbs or
resources. A namespaced `Role` and a `ClusterRole` with the same name must not
be treated as interchangeable.

## Current Behavior

`analyzeRoleBindings` always calls `Roles(rb.Namespace).Get(...)` and does not
honor `rb.RoleRef.Kind`.

This creates two observable errors:

1. If a `RoleBinding` references a `ClusterRole` and no same-named `Role`
   exists in the namespace, the lookup returns `NotFound` and the analyzer
   silently skips the binding. Wildcard permissions in the `ClusterRole` are
   therefore not reported.
2. If a same-named namespaced `Role` does exist, the analyzer inspects that
   unrelated `Role` instead. It can report a false warning even when the
   referenced `ClusterRole` has no wildcard rule, and the warning identifies
   the wrong object kind.

## Steps to Reproduce

1. Create a namespace-scoped `RoleBinding` that references a `ClusterRole`:

   ```yaml
   apiVersion: rbac.authorization.k8s.io/v1
   kind: ClusterRole
   metadata:
     name: namespace-admin
   rules:
   - apiGroups: [""]
     resources: ["pods"]
     verbs: ["*"]
   ---
   apiVersion: rbac.authorization.k8s.io/v1
   kind: RoleBinding
   metadata:
     name: team-binding
     namespace: team-a
   roleRef:
     apiGroup: rbac.authorization.k8s.io
     kind: ClusterRole
     name: namespace-admin
   subjects:
   - kind: ServiceAccount
     name: workload
     namespace: team-a
   ```

2. Run the Security analyzer for that namespace:

   ```shell
   k8sgpt analyze --explain --filter=Security --namespace=team-a
   ```

3. The current analyzer returns no `Security/RoleBinding` result for
   `team-binding`, even though the referenced `ClusterRole` contains a
   wildcard verb.

4. For the inverse case, create a namespaced `Role` named `namespace-admin`
   with a wildcard rule, while changing the `ClusterRole` above to a
   non-wildcard rule. Keep `roleRef.kind: ClusterRole`. The current analyzer
   reports the wildcard from the unrelated `Role` instead of inspecting the
   referenced `ClusterRole`.

## Environment

- **k8sgpt version:** source at `origin/main`, commit `05247a851ba9292ca57e5070f1d0c4d3986b8d4c`
- **Kubernetes version:** any cluster serving `rbac.authorization.k8s.io/v1`
- **AI Backend/Provider:** not involved; this happens before explanation
- **OS/Platform:** reproduced with the repository's Go fake client

## Additional Context

Kubernetes documents that a `RoleBinding` may reference either a namespaced
`Role` or a cluster-scoped `ClusterRole`; a `ClusterRole` referenced by a
`RoleBinding` is still scoped to the namespace of that binding:

- https://kubernetes.io/docs/reference/kubernetes-api/rbac/role-binding-v1/
- https://kubernetes.io/docs/reference/access-authn-authz/rbac/

The source-level fake-client reproduction is deterministic on the baseline:

- ClusterRole-only reference: expected one warning, received zero results.
- Same-name Role plus ClusterRole: expected zero warnings for the referenced
  non-wildcard ClusterRole, received a warning for the namespaced Role.

The focused fix should only correct `RoleBinding` reference resolution and add
both regression cases. It should not expand this issue to `ClusterRoleBinding`
analysis or general RBAC policy auditing.

Collision audit on 2026-08-24 found no matching open or closed Issue/PR for
`RoleBinding ClusterRole`, `wildcard permissions`, or `Security analyzer`.

junnhwan:
Closes #1754

## 📑 Description

The Security analyzer always resolved RoleBinding references through the
namespaced Roles API and ignored roleRef.kind.

This caused ClusterRole references to be silently skipped when no same-named
Role existed, and caused an unrelated same-named Role to be analyzed instead.

This change selects the namespaced Role or cluster-scoped ClusterRole lookup
according to roleRef.kind. It also adds fake-client regressions for both cases.

ClusterRoleBinding and other analyzers are intentionally out of scope.

## ✅ Checks

## ℹ Additional Information

Verified locally:

- `go test ./pkg/analyzer/ -count=1`
- `go test ./... -count=1`
- `go vet ./pkg/analyzer/`
- `go vet ./...`
- `gofmt`
- `git diff --check`

### DIFF
--- a/pkg/analyzer/security.go
+++ b/pkg/analyzer/security.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 
 	"github.com/k8sgpt-ai/k8sgpt/pkg/common"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for keeping the fix on the RoleBinding boundary. I checked that the lookup now follows `roleRef.kind`, and that the tests cover both the ClusterRole-only case and the same-name Role collision. The failure text now identifies the referenced kind too. CI is green; ready for your final look.

PR #1756, opened 2026-08-26
fix: handle empty AI responses and respect configured language

### LINKED ISSUE #1566: [Question]: Details field is empty and some AI responses are in English despite Portuguese configuration
### Checklist

### Affected Components

- [x] K8sGPT (CLI)
- [x] K8sGPT Operator

### K8sGPT Version

0.2.22

### Kubernetes Version

v1.28.10

### Host OS and its Version

_No response_

### Steps to reproduce

I am using the latest version of both the k8sgpt-operator and k8sgpt-ai, which is the CLI that triggers the AI for error recommendations.

1. helm install my-k8sgpt-operator k8sgpt/k8sgpt-operator --version 0.2.22

2. k8sgpt.yaml
apiVersion: core.k8sgpt.ai/v1alpha1
kind: K8sGPT
metadata:
  name: k8sgpt-sample
  namespace: default
 spec:
  ai:
    anonymized: false
    autoRemediation:
      enabled: false
      resources:
      - Pod
      - Deployment
      - Service
      - Ingress
      similarityRequirement: ""
    backOff:
      enabled: false
      maxRetries: 5
    backend: google
    enabled: true
    language: portuguese
    maxTokens: "2048"
    model: gemini-2.5-pro
    secret:
      key: google-api-key
      name: google-api-secret
    topk: "50"
  analysis:
    interval: 30m
  filters:
  - Pod
  - Deployment
  - Ingress
  - Service
  - Security
  repository: ghcr.io/k8sgpt-ai/k8sgpt
  version: v0.4.23   <--------------------------------- latest version k8stp-ai

3. I changed prompts.go to translate to my language portuguese
               I have already tested 4 different versions of k8sgpt-ai with the same result.
              I have rebuilt the Docker image and re-applied the k8sgpt-ai after changing the prompts.go file to confirm that my changes were used.

package ai

const (
    default_prompt = `Você é um analista de Kubernetes.
    Responda EXCLUSIVAMENTE em Português do Brasil (pt-BR). Nunca use outro idioma.

    Explique o erro do Kubernetes abaixo de forma concisa e proponha uma solução passo a passo.
    Se o contexto for insuficiente, forneça uma mensagem padrão em português: "Não foi possível analisar o erro. Verifique os logs do pod para mais detalhes."

    Entrada (entre ---):
    ---
    %s
    ---
    `

    prom_conf_prompt = `Você é um especialista em Prometheus e Kubernetes.
    Responda EXCLUSIVAMENTE em Português do Brasil (pt-BR). Nunca use outro idioma.

    Explique o erro de configuração do Prometheus abaixo e forneça a solução mais provável em passos.
    Se o contexto for insuficiente, forneça uma mensagem padrão em português: "Não foi possível analisar o erro de configuração. Verifique a sintaxe YAML do seu arquivo."

    Entrada (entre ---):
    ---
    %s
    ---
    `

    prom_relabel_prompt = `Você é um especialista em Prometheus e Kubernetes.
    Responda EXCLUSIVAMENTE em Português do Brasil (pt-BR). Nunca use outro idioma.

    Analise as configurações de scrape do Prometheus abaixo e descreva para cada job os labels de serviços e pods, namespaces, portas e containers correspondentes.
    Se o contexto for insuficiente, forneça uma mensagem padrão em português: "Não foi possível analisar as configurações de relabel. Verifique o arquivo de configuração do Prometheus."

    Entrada (entre ---):
    ---
    %s
    ---
    `

    kyverno_prompt = `Você é um especialista em segurança Kubernetes (Kyverno).
    Responda EXCLUSIVAMENTE em Português do Brasil (pt-BR). Nunca use outro idioma.

    Simplifique o aviso do Kyverno abaixo e forneça o comando kubectl mais provável para corrigir.
    Se o contexto for insuficiente, forneça uma mensagem padrão em português: "Não foi possível analisar o aviso do Kyverno. Verifique o PolicyReport para mais detalhes."

    Entrada (entre ---):
    ---
    %s
    ---
    `

    raw_promt = `{"language": "%s","message": "%s","prompt": "%s"}`
)

var PromptMap = map[string]string{
    "raw":                           raw_promt,
    "default":                       default_prompt,
    "PrometheusConfigValidate":      prom_conf_prompt,
    "PrometheusConfigRelabelReport": prom_relabel_prompt,
    "PolicyReport":                  kyverno_prompt,
    "ClusterPolicyReport":           kyverno_prompt,
}

### Expected behaviour

I expect all analyses to return with a populated details field, containing the AI's response in Portuguese, as configured.

### Actual behaviour

Empty Details Field: Many analyses return with the details field in the Result Custom Resource being empty, meaning there is no AI response.

Incorrect Language: Even with my prompts.go file written entirely in Portuguese and the kind: k8sgpt-ai resource configured with the language: portuguese parameter, some analyses are still coming back with the AI's response in English.

I appreciate your help in advance.

### Additional Information

Example: The empty field

"spec": {
    "backend": "google",
    "details": "",  <---------------------------------------------
    "error": [
      {

Example: analysis in English
 Backend:  google
  Details:  Error: The 'file' container has .....................
Solution: 1. Edit the DaemonSet `xxxxxxxx. 2. Navigate to the `file` container spec. 3. Set `file` to `false`.

voltagebots:
Closes #1566

## Description

Improves AI response handling where analysis details come back empty or responses do not consistently use the configured language.

Rebased onto current `main`. The scope is now narrower than when this was first opened, because #1781 landed the empty-choices guard in the meantime. That check is upstream's and is left exactly as it is here. The remaining changes are the parts still unaddressed.

This change now does three things.

Makes the output-language instruction explicit in the default, Prometheus and Kyverno prompts. The previous wording described the requested language but never gave the model a direct instruction, which is the reported symptom in #1566 and in the Korean reproduction added later.

Rejects whitespace-only completions. Upstream now errors when a provider returns zero choices, but a provider returning a single choice containing only whitespace still counts as success and surfaces as an empty `details` field. The error includes the finish reason, which is what distinguishes a truncated response from an empty one.

Fixes two cache faults. A cached entry that decodes to whitespace is treated as a hit and returned as a successful but empty analysis, so it is now treated as a miss and regenerated. And the cache key did not include the prompt template, so changing prompt wording did not invalidate explanations generated under the old wording. Since this PR changes prompt wording, that second fault would otherwise cause existing users to keep receiving stale pre-change output.

## Tests

- `pkg/ai` covers whitespace-only content, including the finish reason, and confirms the non-empty path is unaffected.
- `pkg/analysis` covers the empty-cache-hit fallthrough, that a non-empty hit is still served directly, and that an entry cached under an older prompt template is not served after the wording changes.

`go build ./...`, `go test ./pkg/ai/... ./pkg/analysis/...` both pass on current `main`.

## Note on the underlying issue

#1566 also contains a request from a maintainer for a full Result CR and matching container logs, which the reporters have not supplied. This PR does not close out that diagnostic thread. It fixes the failure modes that are visible in the code regardless of what those logs show, and leaves any backend-specific language contract work separate.


AlexsJones:
Thanks for the focused coverage around empty responses and cache regeneration. The patch is still a draft and EasyCLA is failing, so please sign the CLA and mark it ready when you want the full review.

voltagebots:
> Thanks for the focused coverage around empty responses and cache regeneration. The patch is still a draft and EasyCLA is failing, so please sign the CLA and mark it ready when you want the full review.

Thanks, all done.

voltagebots:
Rebased onto current `main` and narrowed the scope.

#1781 landed the empty-choices guard after this was opened, so that part is now upstream's and I have left it untouched. This branch no longer duplicates it. The diff is smaller as a result.

What remains is the part still unaddressed. The explicit output-language instruction in the three prompts, rejection of whitespace-only completions with the finish reason included, and the two cache faults, namely a whitespace-only cached entry being served as a successful empty analysis, and the prompt template not forming part of the cache key.

That last one matters specifically because this PR changes prompt wording. Without it, existing users would continue receiving explanations generated under the previous prompt and would not see the language fix take effect.

CLA is signed and DCO passes. Ready for review whenever you have time.

AlexsJones (approved):
Thanks for rebasing and narrowing this after #1781, the scope is easy to follow now. The three remaining pieces look right: the explicit `Respond ONLY in the --- %s --- language.` line keeps the `%s` order intact in all three prompts, a whitespace-only completion now errors with its finish reason, and adding `promptTmpl` to the cache key means users actually get the reworded prompts instead of stale cached answers. One small, non-blocking thing: the tests in `cache_path_test.go` use `cache.New(...)`, which falls back to the real file cache, so they leave entries behind in the runner's cache dir. A `t.Cleanup` that removes the seeded keys would keep that tidy. Approving. The test, lint and build runs on this fork head still need a maintainer to approve them before merge.

voltagebots:
Thanks for the review. Both points are addressed.

Added `t.Cleanup` removal for the cache entries the tests seed. It covers slightly more than the seeded keys, because two of the three tests cause the analysis to regenerate and store under a second key, so cleaning only what was seeded would still have left an entry behind. The prompt-versioned test is the clear case, since it seeds under the old template and the run writes a fresh entry under the new one.

Verified rather than assumed. After a run, all four keys the tests touch are absent from the cache directory, and repeated runs leave the entry count flat.

One thing worth flagging honestly, since you may or may not consider it in scope. The same fallback affects existing tests in this package. `analysis_test.go` and `anonymize_test.go` already call `cache.New("test-cache")`, `cache.New("enabled-cache")` and similar, which hit `FileBasedCache` for the same reason and leave entries behind with no cleanup. They use fixed keys so the directory plateaus rather than growing, which is why it is easy to miss. I have deliberately left those alone rather than widen this PR, but happy to follow up separately if you would like them tidied the same way.

Also rebased onto current `main`, since the PR had gone behind again.

AlexsJones (approved):
Thanks for the cleanup, and for checking that the cache entry count stays flat across runs instead of assuming it. Having `cleanupKey` also cover the regenerated `newKey` in the prompt-versioned test is the right call. The rest of the head is the rebase onto #1790, #1792 and #1795, and the PR's own diff is unchanged apart from the test cleanup. A separate PR to tidy the older `analysis_test.go` and `anonymize_test.go` cache usage would be welcome if you feel like it. Approving. The build, test and lint runs on this fork head still need a maintainer to approve them before merge.

PR #1759, opened 2026-08-27
fix: honor StatefulSet start ordinals

### LINKED ISSUE #1757: [Bug] StatefulSet analyzer ignores non-zero start ordinals
## Expected Behavior

The StatefulSet analyzer should inspect the Pods in the ordinal range defined by the StatefulSet: `[spec.ordinals.start, spec.ordinals.start + spec.replicas)`. When `spec.ordinals` is omitted, the start ordinal should remain `0`.

For a StatefulSet with `start: 1` and `replicas: 2`, the analyzer should inspect `database-1` and `database-2`, and report a failure when either Pod is not running.

## Current Behavior

`StatefulSetAnalyzer` always starts its Pod lookup at ordinal `0` and loops through `0..spec.replicas-1`. It therefore looks for `database-0` even when the controller created the valid Pod range `database-1` and `database-2`.

If `database-0` does not exist, the analyzer breaks out before checking the real StatefulSet Pods. A Pending or otherwise unhealthy `database-1` can consequently produce no StatefulSet result. The same zero-based assumption also prevents the existing first-missing-Pod Event fallback from running when the configured first ordinal is non-zero.

## Steps to Reproduce

1. On a Kubernetes cluster where StatefulSet start ordinals are available, create a StatefulSet with a non-zero start ordinal:

   ```yaml
   apiVersion: apps/v1
   kind: StatefulSet
   metadata:
     name: database
     namespace: default
   spec:
     serviceName: database
     replicas: 2
     podManagementPolicy: Parallel
     ordinals:
       start: 1
     selector:
       matchLabels:
         app: database
     template:
       metadata:
         labels:
           app: database
       spec:
         containers:
         - name: database
           image: nginx
   ```

2. During reconciliation, make `database-1` remain Pending while `database-2` is Running. The StatefulSet status will reflect fewer available replicas, for example `availableReplicas: 1`.
3. Run `k8sgpt analyze --filter StatefulSet`.
4. Observe that the current analyzer produces no StatefulSet result because it first requests the non-existent `database-0` and stops. The expected result identifies `database-1` as not running.

The same behavior is reproduced deterministically with the repository fake client using a StatefulSet with `replicas: 2`, `spec.ordinals.start: 1`, `database-1` Pending, `database-2` Running, and `status.availableReplicas: 1`: the baseline at `origin/main` (`05247a8`) returns zero results.

## Environment

- **k8sgpt version:** main (`05247a851ba9292ca57e5070f1d0c4d3986b8d4c`)
- **Kubernetes version:** 1.31+ for the stable StatefulSet start ordinal feature
- **AI Backend/Provider:** n/a; the defect occurs before AI explanation

## Additional Context

Kubernetes defines StatefulSet Pod identity by ordinal. The optional `spec.ordinals.start` field changes the first ordinal, and the assigned range is `[spec.ordinals.start, spec.ordinals.start + spec.replicas)`. The current source in `pkg/analyzer/statefulset.go` does not read `spec.ordinals`.

References:

- https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#start-ordinal
- https://kubernetes.io/docs/reference/kubernetes-api/apps/stateful-set-v1/

A focused fix should derive a default-zero start ordinal from `spec.ordinals.start`, use it for the Pod lookup range, and use it when deciding whether to fetch the first-missing-Pod Event. This issue does not change StatefulSet status semantics, service validation, Event matching, or other analyzers.

The regression test covers both a non-zero ordinal Pod that is not running and a missing first non-zero ordinal with a StatefulSet Warning Event. On the isolated candidate worktree, the fix passes the focused test, all StatefulSet analyzer tests, `go test ./... -count=1`, `go vet ./pkg/analyzer/`, `go vet ./...`, `gofmt`, and `git diff --check`.

No matching open Issue or PR was found for StatefulSet start ordinals, `spec.ordinals`, `StatefulSetStartOrdinal`, or this analyzer behavior. Closed #1249 / PR #1256 concern retrieving StatefulSet Events, not the configured ordinal range.

junnhwan:
Closes #1757

## 📑 Description

The StatefulSet analyzer assumed Pod ordinals always started at 0. For StatefulSets with `spec.ordinals.start` set to a non-zero value, it queried the wrong first Pod, stopped on NotFound, and could miss an unhealthy Pod or the first-missing-Pod Event. This change defaults the start ordinal to 0, checks `[start, start+replicas)`, and uses the same start when deciding whether to fetch the first-missing-Pod Event. It adds two fake-client regression cases. No status, service validation, event identity, or other analyzer behavior is changed.

## ✅ Checks

## ℹ Additional Information

Verified locally on latest `main` (`b5b1a5b`):
- `go test ./pkg/analyzer/ -count=1 -run '^TestStatefulSetAnalyzerUsesConfiguredStartOrdinal$'` — passes
- `go test ./... -count=1 -coverprofile=/tmp/k8sgpt-statefulset-1757.coverprofile` — passes
- `gofmt -l pkg/analyzer/statefulset.go pkg/analyzer/statefulset_ordinal_repro_test.go` — clean
- `git diff --check` — clean

Regression tests added:
- `TestStatefulSetAnalyzerUsesConfiguredStartOrdinal` — covers an unhealthy non-zero ordinal Pod and the missing first non-zero ordinal Event path.


AlexsJones (changes requested):
Thanks for this, and sorry it waited so long. The fix is right: starting the loop at `spec.ordinals.start` and using the same start for the first-missing-Pod event check covers both ways the analyzer went wrong, and the two fake-client cases pin them.

One thing before it can merge: it now conflicts with main. #1790 changed that same event lookup in `statefulset.go` to `util.FetchLatestEvent(a.Context, a.Client, corev1.ObjectReference{Kind, Namespace, Name, UID})`. When you rebase, keep #1790's call and just swap `i == 0` for `i == startOrdinal`. Your second test case should still pass, since the fake StatefulSet has no UID and the event matches on kind and name.

AlexsJones (approved):
Thanks for the clean rebase, that is exactly the shape I was hoping for. #1790's identity-scoped `util.FetchLatestEvent` call survives untouched and the only edit left on that line is `i == 0` becoming `i == startOrdinal`. The loop now walks `[start, start+replicas)`, so a StatefulSet with `ordinals.start: 1` checks `database-1` and `database-2` instead of `database-0` and `database-1`, and `startOrdinal` still defaults to 0 when `spec.ordinals` is nil so existing behaviour is unchanged. Both regression cases pin the two ways it went wrong, and the event case matches on kind and name as expected with no UID on the fake object. All eight checks are green. Approving.

PR #1734, opened 2026-08-12
fix: correct --with-stat flag description

Navneet072300:
## 📑 Description

Split out from #1712 per review feedback (one narrow fix per PR).

The `--with-stat` help text claimed the flag "disables errors display", but `textOutput` always prints the Warnings block regardless of `WithStats`; stats print in addition to the normal output (`cmd/analyze/analyze.go` prints `output_data` unconditionally after `PrintStats`), never as a replacement for it. Confirmed this against current `main` — it's stale documentation, not a missing code change.

## ✅ Checks
- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

## ℹ Additional Information
Text-only change.

PR #1761, opened 2026-08-28
feat: add --resource flag to analyze a single resource

### LINKED ISSUE #1632: Analyze specific deployment
Hello,

Is it possible to analyse a specific deployment via cli?

Tried checking in docs but only got command for scanning a specific namespace. Is there any way to scan a specific deployment or a pod?

Thanks

Bisman-Singh:
Closes #1632

## 📑 Description

Adds `--resource Kind/Name` to `k8sgpt analyze`, so a single workload can be analyzed instead of the whole cluster or namespace:

```
k8sgpt analyze --resource Deployment/my-app -n my-namespace
```

Following the guidance on the issue:

- **Kind selects the analyzer**, reusing the mechanism `--filter` already provides.
- **Name is pushed down to the API server** as a `metadata.name` field selector, via a new `Analyzer.ListOptions()` helper, rather than filtering an already-wide result set client-side. The analyzers that previously built list options inline now share that helper.
- **Scoped with the existing namespace flag** (`-n`), unchanged.
- **Server and MCP modes are untouched.** `NewAnalysis` keeps its current signature and the field is set on the returned config, so this change stays on the CLI path.
- **Behaviour is unchanged when the flag is absent** — `ListOptions()` then returns exactly the label-selector-only options used before.

## ✅ Checks

On documentation: the flag is user-facing, but the README documents usage examples rather than an exhaustive flag list (the closest existing analogue, `--selector`, is not documented there either) and the user docs live in a separate repo. Happy to open a follow-up docs PR if you would like this covered there.

## ℹ Additional Information

**Tests added**

- `pkg/common/listoptions_test.go` — selector construction, including the flag-absent case.
- `pkg/analyzer/resource_selector_test.go` — the selected kind, a name that matches nothing, and the flag absent.

One note worth flagging: the fake clientset **ignores `FieldSelector`** and returns every object, so a naive test here would pass while proving nothing. These tests install a reactor that honours the field selector the way the API server does, and additionally assert the selector that was actually sent on the list call.

**Verified against live clusters (kind)**

Three broken deployments in one namespace:

| Command | Result |
|---|---|
| `analyze --filter Deployment -n demo` | all 3 reported (baseline) |
| `analyze --resource Deployment/alpha-api -n demo` | only `alpha-api` reported |
| `analyze --resource Deployment/does-not-exist -n demo` | `No problems detected`, exit 0 |
| `analyze --resource Deployment` (malformed) | clear error, exit 1 |
| `analyze --resource Pod/<name> -n demo` | only that pod, from 3 |

The Gateway API path was verified separately on a cluster with the Gateway API CRDs installed, since `CtrlClient` is an uncached client and the field selector is served by the API server the same way: with two HTTPRoutes in a namespace, `--resource HTTPRoute/<name>` returned only the named one, and a non-matching name returned nothing.

`go build ./...`, `go vet ./...` and the full `go test ./...` suite (19 packages) pass. `golangci-lint` v2.12.2 (the version CI pins) reports the same pre-existing findings on this branch as on `main` — no new issues.

The branch is built on current `main`.


AlexsJones (changes requested):
Thanks for the focused resource selection work. One thing to fix: when `--resource Deployment/foo` is combined with multiple `--filter` values, `ResourceName` is passed to every selected analyzer, so a `Service` analyzer can also query `metadata.name=foo`. Either reject a filter set that does not match the resource kind, or reduce it to that kind, and cover that combination in the CLI tests.

Bisman-Singh:
Good catch — confirmed and fixed in 07ba4ac.

I reproduced it first on a kind cluster with a Deployment and a Service sharing one name: `--resource Deployment/shared-name --filter Deployment,Service` did indeed report the Service too.

The two flags are now reconciled up front (`resolveResourceSelection` in `cmd/analyze`):

- no `--filter` → the resource kind is selected, as before;
- `--filter` naming the same kind (case-insensitive) → reduced to that kind;
- `--filter` naming any other kind → **rejected** with an error that says how to fix the invocation, e.g. `--resource Deployment/foo selects only Deployment, but --filter also requests "Service"; drop --filter or set it to Deployment`.

I went with reject-on-contradiction rather than silently narrowing, since discarding an explicit `--filter Service` would hide a user mistake — happy to switch to silent reduction if you'd prefer that behaviour.

Covered by CLI tests for the matching, mismatched, multi-kind, case-variant, and malformed combinations plus the no-flag passthrough, and re-verified the rejected/matching paths against a live kind cluster. Full suite still green.

AlexsJones (approved):
Thanks for the quick turnaround, this resolves the flag interaction cleanly. `resolveResourceSelection` rejects a `--resource`/`--filter` mismatch instead of applying the name to unrelated analyzers, and the CLI test plus the field-selector reactor cover the combination I raised. Approving for your final look.

Bisman-Singh:
Thanks for the review — the flag-interaction catch made this a better change.

One heads-up: I rebased onto `main` after your approval (`07ba4ac` → `560a108`) because #1734 landed on `cmd/analyze/analyze.go`, the same file this touches, and the PR had gone `behind`. It's a rebase only — no code change from what you approved. Clean rebase, full suite green, and GitHub carried your approval across to the new head.

Nothing further from my side, happy for this to go in whenever suits.

AlexsJones (approved):
Re-review at the rebased head: the `--resource` flag (kind picks the analyzer, name becomes a field selector, `--filter` and `--resource` kept consistent) is unchanged, just rebased over main. Still approve-recommend.

PR #1766, opened 2026-08-30
feat: prompt to auto-pull missing ollama models

subratojec:
## Description

This PR adds a quality-of-life feature to the Ollama AI backend. Previously, if a user requested a model that wasn't installed locally (e.g. `llama3`), the `Generate` endpoint would immediately fail. 

This change first calls the `List` endpoint to check for the model's existence. If it's not found, it interactively prompts the user for permission `(y/N)` to automatically download the missing model using the `Pull` endpoint before proceeding with generation.

cc @AlexsJones 

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?
- [x] Tested locally by running `k8sgpt analyze --explain --backend ollama --model <missing_model>`
- [x] Verified user prompt appears and correctly triggers model pull on `y`


AlexsJones (changes requested):
Thanks for this, auto-pulling a missing model is a nice touch for the interactive case. Two things before this can land:

1. `GetCompletion` now blocks on `os.Stdin`, which breaks non-interactive use. The server path calls this without a TTY, so it would hang or fail waiting for input. Gate the prompt on stdin being a terminal (or make the pull opt-in behind a flag/env) so server mode and CI are unaffected.
2. The pull progress callback is empty, so a large model looks frozen. Surface the download progress through `ProgressResponse` instead of discarding it.

Worth a quick test that a missing model with a non-TTY stdin returns the error instead of prompting.

subratojec:
Thanks for the feedback, @AlexsJones! Addressed in the latest commit:

1. Env Var Opt-in: Auto-pull now requires K8SGPT_OLLAMA_AUTO_PULL=true. I chose an env var over a CLI flag to keep this Ollama-specific logic isolated without needing to modify IAIConfig for all the other backends.
2. Progress Bar: progressFunc now prints live download progress.
3. Tests: Added a unit test verifying the missing-model error (which also fixes the Codecov warning).

Let me know if this looks good to go!

AlexsJones (approved):
Thanks for reworking this. The `K8SGPT_OLLAMA_AUTO_PULL` opt-in gate (default off) and the populated progress callback both address the review, and the `List` check keeps the error path clean when a model is missing.

One small thing to keep in mind: the progress writes to stdout with carriage returns, which is fine for the opt-in CLI case but worth keeping out of any JSON output path. Good to go from me.

subratojec:
@AlexsJones - I just pushed the quick errcheck linter fix for the mock server! Since it was a new commit, it looks like it paused the auto-merge and the workflows need to be re-approved to run.

subratojec:
Hey @AlexsJones, I’ve just pushed the missing unit tests for the K8SGPT_OLLAMA_AUTO_PULL=true logic to fix the failing codecov/patch check. Could you please approve the workflows to run on this new commit so we can get it merged? Sorry for all the back-and-forth, and thanks!

AlexsJones (changes requested):
Thanks for sticking with this, and sorry your last two pings went unanswered. It showed as merged on our side after the auto-merge was paused.

The workflows have now run on `55539eb`: build, tests and both codecov checks pass. The only failure is golangci-lint on the new tests. In `pkg/ai/ollama_test.go`, lines 25, 66 and 67 call `os.Unsetenv`/`os.Setenv` without checking the error (errcheck). Using `t.Setenv("K8SGPT_OLLAMA_AUTO_PULL", "")` and `t.Setenv("K8SGPT_OLLAMA_AUTO_PULL", "true")` fixes that and restores the variable after each test, so the `defer` can go. Since the gate only checks for `"true"`, the empty value behaves the same as unset. Once lint is green, this is ready for the final look.

subratojec:
"Thanks for bearing with me @AlexsJones! I just pushed the update to use t.Setenv and removed the defer as requested. Should be completely green now—ready for your final look! "

AlexsJones (changes requested):
Thanks for the `t.Setenv` change, that's the right fix and the test file reads clean now. Two mechanical things still block merge: the DCO check fails on `1e27693f` (the new commit has no `Signed-off-by`, so `git commit --amend -s` and push), and the workflow runs on this head still need approving on our side before lint and the tests run. Once the sign-off is on, the workflow approval and a re-check follow.

subratojec:
Ah, sorry @AlexsJones about that! I forgot the sign-off flag on the last commit. Just force-pushed the amended commit with the DCO signature included. Should be good to approve for the final checks now!

AlexsJones (approved):
Thanks for amending the sign-off, DCO is green on `8c349b4` now. The code is the same as the last round: auto-pull stays behind `K8SGPT_OLLAMA_AUTO_PULL=true`, a missing model otherwise fails with a clear message, and the three httptest cases cover missing, pulled and existing models. The test mock still satisfies the full `IAIConfig` interface on current main. Approving. Lint and tests on this head still need a maintainer to approve the workflow runs, and the branch is behind main, so it'll need updating at merge time.

AlexsJones (changes requested):
Thanks for rebasing and squashing this onto current main, the diff is exactly the one I approved on the 17th.

That's also the problem, and it's on me: the build and lint runs never ran on that head, and now that they have, `pkg/ai/ollama_test.go` doesn't compile. Switching to `t.Setenv` removed the last `os.` call, but the `"os"` import on line 8 is still there, so both `build` and `golangci-lint` fail with `"os" imported and not used`.

Drop that one import line and push, and I'll re-approve as soon as CI is green. Sorry for approving before the checks had a chance to run.

subratojec:
My bad, silly mistake leaving that `os` import! I just removed it and force-pushed. It should be good to go once you have a chance to re-approve the CI run. Thanks!

PR #1768, opened 2026-08-31
feat: add ResourceClaim analyzer for missing DeviceClasses

### LINKED ISSUE #1767: feat: add an opt-in ResourceClaim analyzer for missing DeviceClass references
## Is your feature request related to a problem?

Kubernetes Dynamic Resource Allocation (DRA) uses `ResourceClaim` objects to request devices. Each device request references a cluster-scoped `DeviceClass` by name. k8sgpt currently has no analyzer for these resources, so a claim that points to a deleted or never-created `DeviceClass` is not reported by the cluster analysis.

This can leave a workload unable to obtain its requested device while the cause is only visible by manually comparing the claim with the cluster's `DeviceClass` objects.

Kubernetes v1.32 documents DRA as a beta feature that is disabled by default, so this should be an optional analyzer and must not change the default analyzer set.

Add an optional `ResourceClaim` analyzer for the `resource.k8s.io/v1beta1` API:

1. List cluster-scoped `DeviceClass` objects and build the set of available class names.
2. List `ResourceClaim` objects through the existing Kubernetes client.
3. Report each claim that has a `spec.devices.requests[].deviceClassName` reference which does not resolve to an existing `DeviceClass`.
4. Register the analyzer as an additional filter named `ResourceClaim` and document it in the analyzer list.
5. Preserve the existing namespace and label-selector behavior for the claims being analyzed. The label selector must not be applied to the reference lookup, because hiding a `DeviceClass` through a label selector would create a false missing-reference finding.

The regression coverage should include:

- a claim with one valid and one missing `DeviceClass` reference;
- duplicate requests for the same missing class being reported once per claim;
- a claim with no device requests producing no finding;
- label selection applying to claims without hiding referenced `DeviceClass` objects.

## Out of scope

- reporting an unallocated claim as an error, because allocation can legitimately be pending;
- validating driver-specific selectors or device capacity;
- analyzing `ResourceClaimTemplate`, `ResourceSlice`, or Pod-to-claim wiring;
- changing the default analyzer set or enabling the DRA feature gate.

## Additional context

The repository already depends on `k8s.io/api` and `k8s.io/client-go` v0.32.3, which provide the typed `resource.k8s.io/v1beta1` API and fake-client support. The implementation can therefore stay within the existing analyzer contract and does not require a dependency upgrade.

I searched the open and closed k8sgpt issues and pull requests for `ResourceClaim`, `ResourceSlice`, `DeviceClass`, and `Dynamic Resource Allocation` and did not find an overlapping analyzer proposal. The change is limited to the analyzer implementation, focused fake-client tests, analyzer registration, and README documentation.

## Would you be willing to help implement this?

Yes. I would be happy to implement this focused scope with fake-client regression tests and documentation after maintainers confirm the analyzer boundary.

junnhwan:
Closes #1767

## 📑 Description

Adds an optional `ResourceClaim` analyzer for Kubernetes Dynamic Resource Allocation.

- lists cluster-scoped `DeviceClass` objects and checks claim device requests against them;
- reports missing `spec.devices.requests[].deviceClassName` references;
- deduplicates repeated references within one claim;
- keeps the analyzer optional so clusters without the DRA API are unaffected by default;
- applies namespace and label selection to claims without filtering the reference set.

## ✅ Checks

## ℹ Additional Information

Verified locally:

- `go test ./pkg/analyzer -run ResourceClaim -count=1` — passes
- `go test -race ./pkg/analyzer -run ResourceClaim -count=1` — passes
- `go test ./pkg/analyzer -count=1` — passes
- `go build ./...` — passes
- `git diff --check` — clean

Regression tests added:

- missing and valid `DeviceClass` references;
- duplicate missing-class references;
- claim-only label-selector behavior.

The repository-wide golangci-lint invocation still reports five pre-existing findings in `events_test.go`, `httproute.go`, `node.go`, and `ingress_test.go`; none are in the changed files.


AlexsJones (approved):
Looks solid. Keeping `DeviceClass` cluster-scoped and applying the label selector only to claims correctly avoids turning a filtered-out claim into a false missing-reference finding. The `reportedDeviceClasses` dedup and the empty-name skip are both right, and the two tests cover the missing-class and selector-only cases. Optional by default is the right call for the DRA API.

AlexsJones (changes requested):
Thanks for the ResourceClaim analyzer, and sorry this stalled. The auto-merge queued on 09-01 never went through, and GitHub now reports a conflict with main.

Main has moved 15 commits since then, and two changes land on the same spots:
- `pkg/analyzer/analyzer.go`: #1765 added `ValidatingAdmissionPolicy` to `additionalAnalyzerMap` and gofmt re-aligned the whole map, so your one-line registration conflicts. Rebase and re-add the `ResourceClaim` entry.
- #1761 added `--resource`, and analyzers now list with `a.ListOptions()` instead of building `metav1.ListOptions{LabelSelector: a.LabelSelector}` themselves. Please switch the claims list in `resource_claim.go` (line 61) to `a.ListOptions()` so single-resource analysis works for claims. The `DeviceClass` list should keep its unfiltered options, since that's the reference set.

Once it's rebased and green, this is ready for the final look again.

junnhwan:
Thanks for the review! Rebased onto main, restored the ResourceClaim registration, and switched the claims list to a.ListOptions(), keeping the DeviceClass lookup unfiltered. I also added a regression test for resource-name filtering. CI is green and this is ready for another look.

AlexsJones (approved):
Thanks for the rebase and the regression test. `a.ListOptions()` on the claims list keeps the name and label filters where they belong, and the DeviceClass lookup stays unfiltered so a selector can't turn into a false missing-class report; `TestResourceClaimAnalyzerAppliesResourceNameToClaimsOnly` pins exactly that. CI is green on 65869c7. Approving, this is good to merge.

PR #1769, opened 2026-08-31
test: cover the analyzer registry in pkg/analyzer

Bisman-Singh:
Relates to #889

## 📑 Description

`pkg/analyzer` measures 87.5% of statements, but `analyzer.go` had no test file at all, so both registry functions were entirely uncovered. They decide what k8sgpt can do:

- `ListFilters()` backs `k8sgpt filters list` and the values `--filter` accepts
- `GetAnalyzerMap()` decides what actually runs

This adds `pkg/analyzer/analyzer_test.go` covering:

- every registered analyzer is advertised by `ListFilters`
- the core and additional maps neither overlap nor hold `nil` entries
- `GetAnalyzerMap` returns core as core, and merged as a superset of both registries
- the returned maps are copies, not the live package-level registries
- **the filters advertised are exactly the analyzers that can run**

That last one is why I picked this. Registering an analyzer in one map and forgetting the other produces either a filter that advertises an analyzer that never runs, or an analyzer nobody can select — and nothing caught that before.

Integration-provided analyzers depend on activation state, so the assertions cover the core and additional registries only and leave integration behaviour alone.

## ✅ Checks

## ℹ Additional Information

`pkg/analyzer` coverage **87.5% → 89.6%**; `ListFilters` 0% → 70.6% and `GetAnalyzerMap` 0% → 65.0% (the remainder is the integration-activation branches, deliberately not asserted on).

Verified the tests have teeth rather than just passing:

- making `ListFilters` skip one core key fails two tests with `core analyzer "Pod" is registered but not returned by ListFilters`
- making `GetAnalyzerMap` return the live map instead of a copy fails with `coreAnalyzerMap was mutated through the returned map: 0 entries, want 14`

Test-only change; no production code is touched. Full `go test ./...` passes (18 packages).

Scoped deliberately small per the contributing guide. Happy to follow up with per-analyzer tests for the thinner ones (`security.go`, `clusterserviceversion.go`) as separate PRs.

### DIFF
--- a/pkg/analyzer/analyzer_test.go
+++ b/pkg/analyzer/analyzer_test.go
@@ -1,0 +1,171 @@
+/*
+Copyright 2023 The K8sGPT Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package analyzer
+
+import (
+	"sort"
+	"testing"
+)
+
+// keysOf returns the sorted keys of an analyzer map, so failures read as a
+// diffable list rather than as map iteration order.
+func keysOf[V any](m map[string]V) []string {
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
Nice pick, this closes a real gap: `ListFilters` and `GetAnalyzerMap` decide what k8sgpt can do and had no coverage. The overlap, nil-entry, and independent-copy assertions catch the exact failure modes (a filter advertising an analyzer that can't run, or a caller mutating the shared registry), and the sorted `keysOf` helper makes failures diffable. Test-only, no behavior risk. Good to go.

PR #1776, opened 2026-09-05
fix: stop reporting local StorageClasses as deprecated

### LINKED ISSUE #1774: [Bug] Storage analyzer reports local StorageClasses (kubernetes.io/no-provisioner) as deprecated
## Expected Behavior

The Storage analyzer should not report a StorageClass whose provisioner is `kubernetes.io/no-provisioner`. That value is the standard marker for StorageClasses that do not support dynamic provisioning, used with local (static) volumes; it is not a deprecated provisioner.

## Current Behavior

`analyzeStorageClasses` in `pkg/analyzer/storage.go` reports every such StorageClass with:

```
StorageClass <name> uses deprecated provisioner 'kubernetes.io/no-provisioner'
```

The check appears to conflate `no-provisioner` with the in-tree provisioners (`kubernetes.io/gce-pd`, `kubernetes.io/aws-ebs`, ...) that were deprecated as part of CSI migration. `no-provisioner` is a different case: it declares that the StorageClass does not support automatic provisioning and is the documented setup for local volumes, paired with `volumeBindingMode: WaitForFirstConsumer`.

A user with a legitimate local StorageClass therefore gets an error that misattributes their setup as deprecated, and the false positive also consumes an AI explanation.

## Steps to Reproduce

Reproduced deterministically with the repository fake client (no live cluster required):

1. Create a StorageClass with `provisioner: kubernetes.io/no-provisioner`, as in the official example referenced below.
2. Run the Storage analyzer (`k8sgpt analyze --filter Storage`, or the existing `TestStorageAnalyzer` harness).
3. Observe that the analyzer returns one error: `... uses deprecated provisioner 'kubernetes.io/no-provisioner'`.

The existing test case named "Deprecated StorageClass" in `pkg/analyzer/storage_test.go` currently asserts this behavior.

## Environment

- **k8sgpt version:** main (`731a6c9`)
- **Kubernetes version:** n/a (analyzer logic; reproduced with fake client)
- **OS/Platform:** n/a

## Additional Context

Kubernetes documents local StorageClasses with exactly this provisioner; the official example includes the comment "indicates that this StorageClass does not support automatic provisioning":

- https://kubernetes.io/docs/concepts/storage/storage-classes/#local

Suggested fix: remove the `no-provisioner` check and turn the corresponding test case into a regression case asserting no error. A related concern (a local StorageClass set as the cluster default, where PVCs stay Pending) is a separate topic and out of scope here.

No matching open Issue or PR was found for `no-provisioner` or this StorageClass check.

---

*Note: drafted with AI assistance; the referenced code, documentation, and commit history were verified manually.*

bolubo:
Closes #1774

## 📑 Description

The Storage analyzer reports every StorageClass whose provisioner is `kubernetes.io/no-provisioner` as using a deprecated provisioner. That value is the standard marker for StorageClasses that do not support dynamic provisioning, used with local (static) volumes. It is not one of the in-tree provisioners deprecated by CSI migration. A user with a legitimate local StorageClass gets an error that misattributes their setup, and the false positive also consumes an AI explanation.

This change removes that check. A local StorageClass is now reported only for real issues, such as being one of several default StorageClasses.

## ✅ Checks

## ℹ Additional Information

Verified locally:
- `go test ./pkg/analyzer/ -count=1 -run TestStorageAnalyzer` passes
- `go test ./pkg/analyzer/ -count=1` passes
- `go vet ./pkg/analyzer/` clean
- `gofmt` clean
- `go build ./...` clean

The existing Deprecated StorageClass test case was turned into a regression case named Local StorageClass with no-provisioner that asserts no error. It fails on main, where the analyzer still reports the deprecated error, and passes with this change.

Note: drafted with AI assistance; the change and tests were verified locally.

### DIFF
--- a/pkg/analyzer/storage.go
+++ b/pkg/analyzer/storage.go
@@ -68,14 +68,6 @@ func analyzeStorageClasses(a common.Analyzer) ([]common.Result, error) {
 	for _, sc := range scs.Items {
 		var failures []common.Failure
 
-		// Check for deprecated storage classes
-		if sc.Provisioner == "kubernetes.io/no-provisioner" {
-			failures = append(failures, common.Failure{
-				Text:      fmt.Sprintf("StorageClass %s uses deprecated provisioner 'kubernetes.io/no-provisioner'", sc.Name),
-				Sensitive: []common.Sensitive{},
-			})
-		}
-
 		// Check for default storage class
 		if sc.Annotations["storageclass.kubernetes.io/is-default-class"] == "true" {
 			// Check if there are multiple default storage classes
--- a/pkg/analyzer/storage_test.go
+++ b/pkg/analyzer/storage_test.go
@@ -36,17 +36,17 @@ func TestStorageAnalyzer(t *testing.T) {
 		expectedErrors int
 	}{
 		{
-			name:      "Deprecated StorageClass",
+			name:      "Local StorageClass with no-provisioner",
 			namespace: "default",
 			storageClasses: []storagev1.StorageClass{
 				{
 					ObjectMeta: metav1.ObjectMeta{
-						Name: "deprecated-sc",
+						Name: "local-sc",
 					},
 					Provisioner: "kubernetes.io/no-provisioner",
 				},
 			},
-			expectedErrors: 1,
+			expectedErrors: 0,
 		},
 		{
 			name:      "Multiple Default StorageClasses",


AlexsJones (commented):
Thanks for removing the false local-storage deprecation warning. The two-file diff removes only that provisioner check and changes its regression to expect no failure, preserving the other StorageClass checks. Please get the build/analyzer test workflow run on this head before readiness; the current results show validation/DCO/CLA but no build or test result.

bolubo:
Thanks for the review. The Run tests, golangci-lint and Build container checks have not run: GitHub holds workflow runs on fork PRs from first-time contributors in `action_required` until a maintainer approves them, and that approval can only come from the maintainer side. Locally the build and the analyzer suite are green (`go build ./...` and `go test ./pkg/analyzer/...`, including the flipped no-provisioner regression case), so I expect the checks to pass. Could you approve the runs? I will address anything they surface.

PR #1778, opened 2026-09-06
fix: anonymise resource names in event-derived failures

### LINKED ISSUE #560: [Feature] Event Anonymisation
Currently events are not anonymised.
Whilst this is not dangerous in itself, we would prefer to occlude sensitive data wherever possible.
As an example; events may include pod names which might reveal some trade secrets.

This issue is to track and discuss a solution to sending raw event data to the inference AI backends.

trh-ds:
## What

Makes `--anonymize` actually cover the resource names that appear inside Kubernetes event messages.

## Why

Masking in `GetAIResults` is driven entirely by `Failure.Sensitive`:

```go
for _, s := range failure.Sensitive {
    failure.Text = util.ReplaceIfMatch(failure.Text, s.Unmasked, s.Masked)
}
```

Analyzers that build a failure from an event copy the message in and declare nothing sensitive:

```go
// pkg/analyzer/pod.go
if isEvtErrorReason(evt.Reason) && evt.Message != "" {
    failures = append(failures, common.Failure{
        Text:      evt.Message,
        Sensitive: []common.Sensitive{},   // -> sent to the provider verbatim
    })
}
```

So a user who explicitly asked for anonymisation still ships pod, namespace and node names to their AI provider. That is the concern in #560 — *"events may include pod names which might reveal some trade secrets."*

There are roughly 20 call sites passing an empty `Sensitive` today (`pod.go`, `pvc.go`, `job.go`, `statefulset.go`, `rs.go`, `hpa.go`, `configmap.go`, `security.go`, plus the keda/kyverno integrations). #1703 fixed one of them by hand — and even there only the non-event branch, so `statefulset.go` still leaks from its event path. Fixing them one at a time has not converged and leaves every future analyzer to re-introduce the same gap.

## Approach

Derive the masking pairs at the single point every result already passes through, from the result's own identity.

`Result.Name` is the path the analyzer keyed the resource on — `namespace/name`, `namespace/pod/container` for the container-level analyzers, a bare name for cluster-scoped ones like Node. Those segments are exactly the identifiers that appear inside event text, so splitting on `/` gives the right set for free. Analyzers that already declare their own `Sensitive` entries keep working; the two sets compose.

## Also: hardening `ReplaceIfMatch`

It interpolated its pattern straight into a regex:

```go
re := regexp.MustCompile(fmt.Sprintf(`%s(\b)`, pattern))
```

Three problems, all now fixed and covered by table cases:

- Node names are routinely FQDNs, and an unescaped `.` matches any character.
- A value containing `+` or `[` **panics** `MustCompile`.
- An empty pattern compiles to `(\b)`, which matches at every word boundary — `ReplaceIfMatch("new value", "", "X")` returned `"XnewX XvalueX"`.

Node names only begin flowing through this helper as a result of this change, so the escaping is required for the fix to be correct rather than incidental cleanup. The same empty-value hazard exists on the reverse path, where `strings.ReplaceAll` with an empty needle inserts between every character, so that is guarded too.

## Testing

Both fixes were verified to fail without the change.

Reverting the identity masking:

```
--- FAIL: TestGetAIResultsAnonymizesResourceIdentity
    Error: "...--- Readiness probe failed for pod checkout-worker-7d9f in
           namespace prod-payments: connection refused ---..."
           should not contain "checkout-worker-7d9f"
    Messages: pod name was sent to the AI provider despite --anonymize
```

Reverting the `ReplaceIfMatch` hardening:

```
--- FAIL: TestReplaceIfMatch/new_value#01
    expected: "new value"
    actual  : "XnewX XvalueX"
--- FAIL: TestReplaceIfMatch/abc
    expected: "abc"
    actual  : "X"
```
(plus a panic on the `[` case)

The new test asserts on the prompt handed to the AI client, so it checks what would actually leave the machine, and also asserts the masking is reversed so the user still reads real names in the output.

With the change:

```
$ go test ./pkg/analysis/... ./pkg/util/... ./pkg/analyzer/...
ok  	github.com/k8sgpt-ai/k8sgpt/pkg/analysis	12.789s
ok  	github.com/k8sgpt-ai/k8sgpt/pkg/analyzer	9.179s
```

`pkg/util` has one unrelated pre-existing failure on Windows only (`TestEnsureDirExists` asserts a Unix `mkdir` error string); it fails identically on a clean checkout of `main` and passes on the Linux CI runner.

## Scope

Deliberately not attempting to mask image references or registry paths out of free-text event messages. Extracting those reliably needs real parsing, and feeding values containing `.`/`+`/`:` through the redaction helper is what makes the regex issues above dangerous. #560 asks specifically about resource names, which this covers.

Closes #560

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WVmaNMnAydzyhpPwcjPavH

### DIFF
--- a/pkg/analysis/analysis.go
+++ b/pkg/analysis/analysis.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


AlexsJones (changes requested):
Thanks for covering the outgoing prompt as well as the displayed result. Please handle overlapping identifiers before this lands: `maskText` masks namespace segments before resource names, so `Name: prod/api-prod` changes the `prod` suffix in `api-prod` first and then cannot match the full pod name, leaving `api-` visible. The same issue affects declared values that contain a derived identifier. Build one deduplicated mapping and replace complete longest matches without remasking replacements, then add overlapping namespace/resource and declared/derived prompt-redaction tests.

trh-ds:
@AlexsJones thanks — you were right on both counts, and the second one was worse than I realised. Pushed 1bbd44f.

**One deduplicated mapping.** `sensitiveMapping` now merges the declared pairs and the derived identity into a single set: empty values dropped, deduplicated on the unmasked value, ordered longest first. Deduplication also fixes something I had not considered — an identifier that appeared in both sources previously got two different masks, and only whichever the unmasking loop reached first was reversed.

**Complete longest matches, no remasking.** `replaceLongestFirst` builds one `QuoteMeta`-escaped alternation and rewrites in a single left-to-right pass. Leftmost matching handles the case where the shorter value is a *suffix* (`prod` inside `api-prod`); the longest-first ordering handles it as a *prefix* (`prod` inside `prod-api`), where both alternatives start at the same position and Go's alternation preference decides. A single pass means a replacement is never itself rewritten by a later pair.

`GetAIResults` builds that mapping once per result and uses it for masking and unmasking, so overlaps resolve identically in both directions. `unmaskText` re-sorts longest-first on the mask side and drops the trailing `\b`, since base64 masks can end in `=`.

**Tests.** Four new, each verified to fail against the previous implementation:

- `TestGetAIResultsAnonymizesOverlappingNamespaceAndResource` — your exact case. Before the fix the prompt contained `pod api-elJ6Tw== in namespace elJ6Tw==`.
- `TestGetAIResultsAnonymizesDeclaredValueContainingDerivedOne` — an Ingress in `prod/api` referencing service `billing-prod`. Before the fix `billing-` reached the prompt. Worth noting the first shape I tried here did *not* leak: when the declared value''s stem is itself a derived segment it gets masked anyway, just split across two masks. The leak needs a stem nothing else covers, which is what this test uses.
- `TestMaskTextReplacesCompleteLongestMatch` — prefix and suffix nesting at the replacement level, asserting a single whole mask and a clean round-trip.
- `TestSensitiveMapping` — dedup precedence, empty-value handling, ordering.

**One correction to the PR description.** `pkg/analysis` no longer calls `util.ReplaceIfMatch`; the escaping rationale moved into `replaceLongestFirst`. I kept the `ReplaceIfMatch` hardening anyway — it is exported, still called elsewhere, and a value containing `+` or `[` panics `MustCompile` — but it is now a latent-panic fix rather than something this change depends on. Happy to split it into its own PR if you would prefer this one stay narrow.

`go build ./...`, `go vet ./pkg/...` and `gofmt` are clean; `pkg/analysis` and `pkg/analyzer` pass.

AlexsJones (commented):
Thanks for addressing both overlap cases. The combined deduplicated mapping and single-pass longest-match replacement fix the namespace/resource and declared/derived leaks from my review, and the new tests inspect the outgoing prompt. Please update the branch, resolve DCO and attach a current-head test run before the final review; those checks are still pending here.

AlexsJones (changes requested):
Thanks for addressing both overlap cases. The combined deduplicated mapping and single-pass longest-match replacement fix the namespace/resource and declared/derived leaks from my review, and the new tests inspect the outgoing prompt. Please update the branch, resolve DCO and attach a current-head test run before the final review; those checks are still pending here.

trh-ds:
Branch updated and DCO resolved: rebased onto `main` (`e69dbc7`, includes the grpc security bump) and both commits now carry `Signed-off-by`. Head is `a1bf5e5`; DCO, EasyCLA and License Compliance are green, `validate` is queued on the runner.

Test run at that head:

```
$ go build ./...
$ go test ./pkg/analysis/... ./pkg/util/... ./pkg/analyzer/...
ok      github.com/k8sgpt-ai/k8sgpt/pkg/analysis        12.899s
ok      github.com/k8sgpt-ai/k8sgpt/pkg/analyzer        8.963s
--- FAIL: TestEnsureDirExists/#00
    Error: "mkdir : The system cannot find the path specified."
           does not contain "mkdir : no such file or directory"
FAIL    github.com/k8sgpt-ai/k8sgpt/pkg/util            5.296s
```

That `pkg/util` failure is the pre-existing Windows-only one noted in the description — it asserts a Unix `mkdir` error string and is untouched by this branch. Re-confirmed on a clean worktree of `origin/main` at `e69dbc7`:

```
$ go test ./pkg/util/... -run TestEnsureDirExists
--- FAIL: TestEnsureDirExists/#00
    Error: "mkdir : The system cannot find the path specified."
           does not contain "mkdir : no such file or directory"
FAIL    github.com/k8sgpt-ai/k8sgpt/pkg/util            4.892s
```

The redaction tests from your review:

```
$ go test ./pkg/analysis/... -run 'Anonymize|Mask' -v
--- PASS: TestGetAIResultsAnonymizesResourceIdentity
--- PASS: TestGetAIResultsAnonymizesOverlappingNamespaceAndResource
--- PASS: TestGetAIResultsAnonymizesDeclaredValueContainingDerivedOne
--- PASS: TestMaskTextReplacesCompleteLongestMatch
    --- PASS: .../shorter_value_is_a_suffix
    --- PASS: .../shorter_value_is_a_prefix
ok      github.com/k8sgpt-ai/k8sgpt/pkg/analysis        13.098s
```

Ready for the final review.

AlexsJones (approved):
Thanks for the dedup plus longest-first rewrite. The overlap cases I flagged are now pinned by the new tests, and quoting the patterns in `ReplaceIfMatch` also fixes the FQDN and literal-value cases. Ready for your final look.

PR #1782, opened 2026-09-11
fix: do not report local StorageClass (no-provisioner) as deprecated

### LINKED ISSUE #1774: [Bug] Storage analyzer reports local StorageClasses (kubernetes.io/no-provisioner) as deprecated

aoright:
Fixes #1774

The Storage analyzer incorrectly reports StorageClasses with provisioner `kubernetes.io/no-provisioner` as deprecated. This PR removes this check and updates the corresponding test case, as `kubernetes.io/no-provisioner` is the standard and correct way to configure local static storage in Kubernetes.


AlexsJones:
Thanks for the change. This duplicates #1779, which removes the same no-provisioner deprecation check. I've flagged it for the maintainer to decide which to keep, so you may want to coordinate there.

AlexsJones:
Thanks for the fix, @aoright, and for keeping the branch rebased. #1779 by @bolubo makes the same change (removing the no-provisioner deprecation check) and got in first, so closing this as a duplicate to keep a single PR for #1774. Your independent confirmation of the bug is appreciated and noted here. Please keep sending fixes.

PR #1783, opened 2026-09-11
fix: filter service events by involved object kind

### LINKED ISSUE #844: [Question]:    Service analyzer attributes Warning events from same-named non-Service objects to the Service
### Checklist

### Affected Components

- [ ] K8sGPT (CLI)
- [x] K8sGPT Operator

### K8sGPT Version

k8sgpt:v0.4.32

### Kubernetes Version

1.37.0

### Host OS and its Version

_No response_

### Steps to reproduce

## Description
    
    The Service analyzer (`pkg/analyzer/service.go`) lists Warning events by object
    **name only**, without filtering on `involvedObject.kind`. Any Warning event on
    any object in the namespace that shares the Service's name is attributed to the
    Service as a failure.
    
    In a standard operator install this is not a coincidence but structural: the
    operator creates the server's Deployment **and Service** with the name of the
    `K8sGPT` CR (operator `pkg/resources/k8sgpt.go`, `GetService`: `Name: config.Name`).
    So the server Service **always** shares its name with the K8sGPT CR. When the
    Analyze RPC fails (e.g. during a server rollout), the operator emits
    `AnalysisFailed` **Warning** events on the **K8sGPT CR** object, and the Service
    analyzer picks them up:
    
    ```
    Service k8sgpt/ollama has event failed to call Analyze RPC: rpc error:
    code = Unavailable desc = error reading from server: EOF
    ```
    
    The Service is healthy (`1/1` endpoints ready); the reported "failure" is the
    operator's own RPC error.
    
    ## Confusing symptom: deleting the Result does not clear it
    
    Because the analyzer re-derives the failure from the still-present events on
    every pass, deleting the `Result` CR does not help: on the next reconcile (~1
    min) the operator re-creates the Result with the same text. It looks like a
    previously fixed error "came back", but it is not a cache — it re-appears until
    the Warning events expire (kube-apiserver event TTL; default 1h on this
    cluster's apiserver).
    
    ## Reproduction (all observed on the cluster described below)
    
    1. Standard operator install: CR `ollama` → server Deployment + Service `ollama`.
    2. Restart the server Deployment. During the rollout window the Analyze RPC is
       briefly unavailable; the operator records `AnalysisFailed` Warning events on
       the K8sGPT CR (observed: `error reading from server: EOF` and
       `dial tcp <svc-ip>:8080: connect: connection refused`).
    3. Next analysis pass produces a `Result` of kind `Service` containing the
       operator's RPC error text (analyzer source below).
    4. `kubectl delete result <name>` → the Result is re-created on the next
       reconcile with the same text, until the events expire.
    
    ## Root cause (v0.4.32 source)
    
    ```go
    // pkg/analyzer/service.go, func (ServiceAnalyzer) Analyze
    // line 117:
    FieldSelector: "involvedObject.name=" + ep.Name,
    // ...
    // line 126:
    Text: fmt.Sprintf("Service %s/%s has event %s", ep.Namespace, ep.Name, event.Message),
    ```
    
    `involvedObject.name=<name>` matches every object with that name in the
    namespace. The events being attributed in the reported case are on the K8sGPT
    CR (`involvedObject.kind=K8sGPT`); the Service object itself had no Warning
    events. The same code path would also attribute Warning events of any other
    same-named object (e.g. a Deployment or Pod sharing the Service's name) to the
    Service.
    
    ## Proposed fix
    
    Filter by kind as well. `involvedObject.kind` is a supported Events field
    selector (verified on this cluster):
    
    ```go                                                                                                    
    metav1.ListOptions{                                                                                      
        FieldSelector: "involvedObject.kind=Service,involvedObject.name=" + ep.Name,                         
    }                                                                                                        
    ```                                                                                                      
                                                                                                             
    or post-filter with `event.InvolvedObject.Kind != "Service"`. Since the                                  
    misattributed events are raised on non-Service objects, this changes behavior                            
    only for the false-positive case.                                                                        
                                                                                                             
    ## Workaround (until fixed)                                                                              
                                                                                                             
    Delete the stale Warning events; the Result is then auto-removed by the                                  
    operator's stale-result cleanup on the next reconcile:                                                   
                                                                                                             
    ```bash                                                                                                  
    kubectl delete events -n <ns> --field-selector type=Warning                                              
    ```                                                                                                      
                                                                                                             
    ## Environment                                                                                           
                                                                                                             
    - k8sgpt server: `ghcr.io/k8sgpt-ai/k8sgpt:v0.4.32`                                                      
    - k8sgpt-operator: `ghcr.io/k8sgpt-ai/k8sgpt-operator:v0.2.29`                                           
    - Cluster: KIND, Kubernetes v1.37.0 (rootless podman)                                                    
    - kube-apiserver `--event-ttl` default 1h0m0s (not overridden)  

### Expected behaviour

- The Service analyzer should only report events whose `involvedObject` **is                             
      the Service itself** (`kind=Service`, matching name). Warning events raised                            
      on other objects that merely share the name — in this case the operator's own                          
      `AnalysisFailed` events on the `K8sGPT` CR (`kind=K8sGPT`) — should not be                             
      attributed to the Service.                                                                             
    - In the reproduction scenario the Service is healthy (all endpoints ready) and                          
      has no Warning events of its own, so the analysis should produce **no failure                          
      for it**, and no `Result` containing the operator's RPC error text should be                           
      created — nor re-created after the Result is deleted.                                                  
                                                                                                             
    This is a request about event **attribution** in the analyzer; it is not a                               
    request to stop the operator from emitting `AnalysisFailed` events on the                                
    K8sGPT CR.         

### Actual behaviour

_No response_

### Additional Information

_No response_

karangupta982:
Closes [#844](https://github.com/k8sgpt-ai/k8sgpt-operator/issues/844)

## 📑 Description

The service analyzer previously queried kubernetes events using only involvedObject.name, which could cause warning events from other resources with the same name to be incorrectly attributed to a service.

This change adds involvedObject.kind=Service to the event field selector.

A regression test was added to verify that:

- Warning events belonging to same-named non-Service resources are ignored.
- Warning events belonging to the Service are still reported.

## ✅ Checks

* [x] My pull request adheres to the code style of this project
* [ ] My code requires changes to the documentation
* [ ] I have updated the documentation as required
* [x] All the tests have passed

## ℹ Additional Information

No breaking changes or new dependencies are introduced.

### DIFF
--- a/pkg/analyzer/service.go
+++ b/pkg/analyzer/service.go
@@ -116,7 +116,7 @@ func (ServiceAnalyzer) Analyze(a common.Analyzer) ([]common.Result, error) {
 		// fetch event
 		events, err := a.Client.GetClient().CoreV1().Events(a.Namespace).List(a.Context,
 			metav1.ListOptions{
-				FieldSelector: "involvedObject.name=" + ep.Name,
+				FieldSelector: fmt.Sprintf("involvedObject.kind=%s,involvedObject.name=%s", kind, ep.Name),
 			})
 
 		if err != nil {
--- a/pkg/analyzer/service_test.go
+++ b/pkg/analyzer/service_test.go
@@ -23,7 +23,9 @@ import (
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
 )
 
 func TestServiceAnalyzer(t *testing.T) {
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Nice catch on the misattribution, same-named events from other kinds landing on the Service is exactly the kind of quiet wrongness that erodes trust in output.

The one-line selector fix reuses the existing `kind` variable so it cannot drift, and the new test is honest about the fake clientset not filtering (the reactor emulates the API server) with assertions on both the selector string and the attributed failure. Ready for the maintainer's final look.

PR #1790, opened 2026-09-16
fix: scope analyzer events to object identity

### LINKED ISSUE #1789: FetchLatestEvent can attribute same-name events to the wrong object
## Expected behavior

Analyzers should only use Events whose `involvedObject` identifies the resource being analyzed. Identity should include the object kind and, when available, UID in addition to namespace and name.

## Current behavior

`pkg/util.FetchLatestEvent` lists Events with only:

```text
involvedObject.name=<name>
```

The helper is shared by the Pod, Job, PersistentVolumeClaim, StatefulSet, and KEDA ScaledObject analyzers. Kubernetes object names are unique only within a resource kind, so a Pod and Job (or another namespaced object) may legitimately have the same name. The helper then chooses the newest Event across all matching kinds. It can therefore:

- replace a Job's `BackoffLimitExceeded` detail with a same-name Pod Event;
- report a same-name object's Event for a pending PVC, Pod, StatefulSet, or ScaledObject;
- use a stale Event after an object is deleted and recreated with the same name but a new UID.

## Reproduction

On current `main` (`ed1aac46`), I added a fake-client regression with three Events in `default`:

1. `Job/worker`, UID `job-uid`, timestamp 1;
2. `Pod/worker`, UID `pod-uid`, timestamp 2;
3. stale `Job/worker`, UID `old-job-uid`, timestamp 3.

Calling the current helper for the live Job returns `pod-event` rather than `job-event` (the stale same-kind case is newer still). The focused test fails deterministically:

```text
expected: "job-event"
actual:   "pod-event"
```

No live cluster or AI backend is required.

## Proposed scope

Pass a `corev1.ObjectReference` to `FetchLatestEvent`, send `involvedObject.name`, `involvedObject.kind`, and non-empty `involvedObject.uid` together as the Kubernetes field selector, and defensively filter the returned list by the same identity. Update the five callers and their fixtures without changing analyzer result formats or reason handling.

A local patch makes the identity regression pass and keeps the util, analyzer, and KEDA analyzer packages green. I can open the focused PR with those tests.

This is separate from #1752/#1753: that issue is limited to the Service analyzer's direct Event list, while these call sites use the shared helper.

git-jxj:
Closes #1789

## 📑 Description

FetchLatestEvent previously filtered only by involvedObject.name. Kubernetes allows different kinds to share a name, and recreated resources can retain events for an older UID, so analyzers could report an unrelated event.

This change passes the full object reference from each analyzer and scopes event lookup by name, kind, and UID. It also defensively applies the same identity check to returned events, which keeps behavior correct for clients that do not enforce field selectors.

A regression test covers both a same-name Pod event and a stale Job event with an older UID.

## ✅ Checks

Validated with:

- go test ./... -count=1
- go test -race ./pkg/util ./pkg/analyzer ./pkg/integration/keda -count=1
- go vet ./...
- golangci-lint run --timeout=5m --new-from-rev=origin/main ./...

## ℹ Additional Information

No dependencies or user-facing configuration are changed.


AlexsJones (approved):
Nice catch on the same-name collision, and the three-event regression test pins it exactly. Scoping `FetchLatestEvent` by kind, name and UID (plus the defensive re-check for clients that ignore field selectors) is the right fix, and all five callers pass their analyzer's own `kind`, which matches what Kubernetes stamps on `involvedObject.kind`. The fixture updates look complete. Approving. The build, test and lint runs on this fork head are waiting on a maintainer to approve them, so those need to go green before merge.

git-jxj:
Thanks for the review. I re-checked the full PR (including all five callers) and ran the following locally:

- `go test ./... -count=1`
- `go test -race ./pkg/util ./pkg/analyzer ./pkg/integration/keda -count=1`
- `go vet ./...`
- `golangci-lint run --new-from-rev=ed1aac4`
- `git diff --check`

All passed. The remaining Build, Tests, and golangci-lint workflow runs are still `action_required` and need maintainer approval; no additional code changes were needed from this review.

git-jxj:
Synced this branch with the current `main` via the GitHub update-branch flow. The existing approval remains valid and Semantic Validation, DCO, EasyCLA, and License Compliance are green. The newly created Build container, Run tests, and Run golangci-lint workflows are awaiting maintainer approval.

git-jxj:
The refreshed Build, Tests, golangci-lint, Codecov, and compliance checks are now all green. The PR remains approved and conflict-free, so it is ready for maintainer merge.

PR #1792, opened 2026-09-16
fix: send involvedObject filters as a field selector in MCP list-events

### LINKED ISSUE #1784: bug: MCP list-events filters events after the API limit
## Expected Behavior

When the MCP `list-events` tool receives `involvedObjectName` and/or `involvedObjectKind`, those filters should be sent to the Kubernetes API together with the existing list limit. Unrelated events should not consume the first unfiltered page before the tool applies the requested object filter.

The tool already documents both filters, and the built-in `troubleshoot-pod` prompt explicitly tells the agent to use `list-events` with the Pod name.

## Current Behavior

On main at `ed1aac46d79b13d4c92dbc8bdf07a8ae47103073`, `handleListEvents` sends only `Limit` to `Events(namespace).List`. It then filters the returned page locally by object name and kind.

If that page contains only unrelated events, the tool returns `[]`, even when a matching event exists beyond the unfiltered page. This can remove useful evidence from an agent's troubleshooting flow.

## Steps to Reproduce

I reproduced this with the real handler and client-go against an `httptest` Kubernetes API fixture; no live cluster or model is required.

1. Have the API return unrelated events before events for Pod `default/target`. The regression fixture also includes a Deployment named `target` and a Pod with a different name, so both filter conditions are tested independently.
2. Invoke the MCP tool with:

   ```json
   {
     "name": "list-events",
     "arguments": {
       "namespace": "default",
       "involvedObjectName": "target",
       "involvedObjectKind": "Pod",
       "limit": 1
     }
   }
   ```

3. The baseline sends an event-list request with `limit=1` and no `fieldSelector`. The fixture returns the first unrelated event and a continuation token. The handler filters that event out and returns `[]`.
4. Sending `involvedObject.name=target` and `involvedObject.kind=Pod` in the same API request returns the matching event in this scenario.

The minimal triggering pattern is one unrelated event in the first page and one matching event after it. With the default limit, the same issue occurs when the first 100 unfiltered events do not include the target.

Local regression results on the baseline versus a focused patch:

| Case | Baseline | Patched |
| --- | --- | --- |
| Name and kind, limit 1 | Incorrect empty array | Matching event |
| Name only / kind only | Incorrect empty array | Matching event |
| Multiple matches / all namespaces | Incorrect empty array in the fixture | Requested matches |
| No filters / no matches / default and capped limits | Pass | Pass |
| Kubernetes API 403 remains a tool error | Pass | Pass |

## Environment

- **k8sgpt version:** main, `ed1aac46d79b13d4c92dbc8bdf07a8ae47103073`
- **Kubernetes version:** local HTTP fixture using core/v1 Events and client-go v0.32.3; no live cluster tested
- **AI Backend/Provider:** not involved; this MCP tool reads Kubernetes events directly
- **OS/Platform:** Windows amd64, Go 1.26.5

## Additional Context

- [Current handler](https://github.com/k8sgpt-ai/k8sgpt/blob/ed1aac46d79b13d4c92dbc8bdf07a8ae47103073/pkg/server/mcp_handlers.go#L340)
- [Existing tool schema](https://github.com/k8sgpt-ai/k8sgpt/blob/ed1aac46d79b13d4c92dbc8bdf07a8ae47103073/pkg/server/mcp.go#L265)
- [Pod troubleshooting prompt](https://github.com/k8sgpt-ai/k8sgpt/blob/ed1aac46d79b13d4c92dbc8bdf07a8ae47103073/pkg/server/mcp_prompts.go#L39)
- [client-go's Event field selector helper](https://github.com/kubernetes/client-go/blob/v0.32.3/kubernetes/typed/core/v1/event_expansion.go#L131)

I checked the actual patches in #1589 and #1613: the first introduced this ordering, and the latter retained it while adding limit bounds and improving error handling. #1752/#1753/#1783 concern Service-analyzer event identity, rather than this MCP handler's filter/limit order. I did not find an existing issue or PR for this specific behavior in the open/closed searches.

The proposed change is limited to building the two optional field selectors and passing them to the existing `Events.List` call, plus regression tests. It preserves the current event-array response, namespace handling, default/maximum limit, and tool-error behavior.

This does not add automatic pagination or claim that every Kubernetes response fills its requested limit: the Kubernetes API allows a short or empty page with a continuation token. A new pagination interface or result envelope would be a separate discussion.

Would a focused fix along these lines be welcome? I have a local patch and deterministic HTTP-based regression tests ready for review after agreement on the scope.

pujitha24:
Closes #1784

## 📑 Description

The MCP `list-events` tool's `handleListEvents` sent only `Limit` to `Events(namespace).List`, then filtered the returned page locally by `involvedObjectName`/`involvedObjectKind`. If that page contained only unrelated events, the tool returned `[]` even when a matching event existed beyond the unfiltered page — silently dropping evidence from an agent's troubleshooting flow (e.g. the built-in `troubleshoot-pod` prompt, which calls `list-events` with a Pod name).

This builds the `involvedObject.name`/`involvedObject.kind` field selectors (mirroring client-go's own `Events.GetFieldSelector` helper) and passes them on the same `List` call, so the Kubernetes API applies the filter before `Limit` truncates the page. The existing limit bounds (`DefaultListLimit`/`MaxListLimit`) and the 403/tool-error handling are unchanged. The now-redundant local filtering loop was removed since the API performs the filtering.

## ✅ Checks

## ℹ Additional Information

Validation:
- Added `pkg/server/mcp_handlers_test.go`, a regression test that spins up an `httptest`-backed fake Kubernetes API (via a temp kubeconfig, same pattern as `pkg/kubernetes/kubernetes_test.go`) whose events endpoint filters by the received `fieldSelector` query parameter the way a real API server would. The dataset places an unrelated event before the matching one, reproducing the issue's repro shape.
- Confirmed this is a genuine fails-before/passes-after regression test: with only `pkg/server/mcp_handlers.go` reverted to the pre-fix local-filtering code (test file kept), `go test ./pkg/server/... -run TestHandleListEventsAppliesInvolvedObjectFieldSelector -v` fails 3 of 4 subtests with empty results; with the fix restored, all 4 pass.
- Ran `go build ./...` and `go test ./...` (full suite) — both pass.
- Ran `golangci-lint run ./pkg/server/...` (CI pins v2.12.2, local v2.13.2) — only 4 pre-existing `staticcheck` issues remain, all in unrelated files (`query.go`, `server.go`, `server_test.go`); CI runs with `only-new-issues: true` so these are unaffected.
- No live cluster was used; the httptest fixture is a deterministic stand-in for the Kubernetes API, consistent with how the issue itself was reproduced.


AlexsJones (approved):
Thanks for turning #1784 into a clean fix. Sending `involvedObject.name` and `kind` as a field selector means `Limit` now truncates after filtering, which is the actual bug, and the httptest fixture applies the selector before the limit the way the API server does, so the name-only, kind-only and unfiltered cases are all pinned. CI is green. Approving. One merge-order note: #1788 moves `pkg/server` to mcp-go v1, so whichever of the two lands second should get a quick rebuild.

PR #1795, opened 2026-09-17
fix: scope service events to endpoint namespace

### LINKED ISSUE #1752: [Bug] Service analyzer attributes warning Events to the wrong object

git-jxj:
Closes #1752

Supersedes the remaining namespace-scoping part of #1753. The kind/name selector from that report is already present on `main` through #1783; this PR keeps that selector intact and carries only the still-missing namespace fix forward on the current codebase.

## 📑 Description

When the analyzer runs across all namespaces, it lists Endpoints cluster-wide. The Service analyzer then queried Events with `Events(a.Namespace)`. If `a.Namespace` was empty, a warning Event for a same-named Service in another namespace could be attributed to the Endpoint currently being analyzed.

Query Events from `ep.Namespace` so namespace, kind, and name all identify the Service. Add a regression with `team-a/api` and a same-named `team-b/api` Event that verifies only the event from the Endpoint namespace is reported.

## ✅ Checks

- [x] My code does not require documentation changes

Validation performed locally:

- `go test ./... -count=1`
- `go test -race ./pkg/analyzer -run TestServiceAnalyzer -count=1`
- `go vet ./...`
- `golangci-lint run --timeout=5m --new-from-rev=origin/main ./...`
- `git diff --check`

## ℹ Additional Information

No result-format, reason-handling, API, or dependency changes.


git-jxj:
@AlexsJones this is the current-`main` replacement for the remaining namespace-scoping part of #1753, preserving the kind/name selector already merged in #1783. The focused and full local validations listed above pass. When the fork workflows are approved, I would appreciate a review.

AlexsJones (approved):
Thanks for carrying this forward onto current main and keeping #1783's kind/name selector intact. Listing Events from `ep.Namespace` is the right fix: during a cluster-wide run `a.Namespace` is empty, so a same-named Service's warning in another namespace could get attached to the wrong Endpoint. The new `team-a/api` vs `team-b/api` test would fail without the one-line change, so it pins the fix. When `a.Namespace` is set, the Endpoints are listed from that same namespace, so nothing changes there. Build and lint are green. The branch is behind main, but #1790 and #1792 touched different files. Good to merge.

PR #1797, opened 2026-09-22
fix: report blocked PDBs whose selector is empty or uses matchExpressions

### LINKED ISSUE #1796: bug: PDB analyzer ignores blocked PodDisruptionBudgets with empty or matchExpressions selectors
## Expected Behavior

A `PodDisruptionBudget` with `DisruptionAllowed=False` should be reported by the PDB analyzer regardless of which selector shape it uses, since a blocked PDB is what wedges a `kubectl drain` and stalls a node upgrade.

## Current Behavior

The PDB analyzer only reports a blocked PDB when its selector uses `matchLabels`. A blocked PDB is silently dropped when its selector is:

- **empty (`{}`)** — which, per the `policy/v1` contract, selects *every pod in the namespace*, giving it the widest blast radius of any PDB
- **`matchExpressions`-only** — an ordinary, valid selector

In both cases `k8sgpt analyze` reports `No problems detected` on a cluster where evictions are blocked.

The cause is in `pkg/analyzer/pdb.go`: the only place a failure is ever constructed is inside the loop over `Selector.MatchLabels`, so any selector without `matchLabels` produces zero failures.

```go
if pdb.Spec.Selector != nil && pdb.Spec.Selector.MatchLabels != nil {
    for k, v := range pdb.Spec.Selector.MatchLabels {
        failures = append(failures, ...)   // only failure path
    }
}
```

For reference, the `policy/v1` API contract (`k8s.io/api/policy/v1/types.go`) states:

> A null selector will match no pods, while an empty ({}) selector will select all pods within the namespace.

(A **null** selector matching no pods is correctly not reported — it cannot block an eviction. Only the empty and `matchExpressions` cases are wrong.)

## Steps to Reproduce

1. Create a namespace with a single pod and two blocked PDBs:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: pdb-demo
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: pdb-demo
spec:
  replicas: 1
  selector:
    matchLabels: { app: web }
  template:
    metadata:
      labels: { app: web }
    spec:
      containers:
        - name: app
          image: registry.k8s.io/pause:3.10
---
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: blocks-everything
  namespace: pdb-demo
spec:
  minAvailable: 1
  selector: {}                 # selects EVERY pod in the namespace
---
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: expr-selector
  namespace: pdb-demo
spec:
  minAvailable: 1
  selector:
    matchExpressions:
      - { key: app, operator: In, values: ["web"] }
```

2. Confirm both PDBs are genuinely blocking evictions:

```
$ kubectl get pdb -n pdb-demo
NAME                MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
blocks-everything   1               N/A               0                     8s
expr-selector       1               N/A               0                     8s

$ kubectl get pdb -n pdb-demo -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.status.conditions[*].type}{"="}{.status.conditions[*].status}{"\t"}{.status.conditions[*].reason}{"\n"}{end}'
blocks-everything	DisruptionAllowed=False	InsufficientPods
expr-selector	DisruptionAllowed=False	InsufficientPods
```

3. Run the analyzer:

```
$ k8sgpt analyze --filter PodDisruptionBudget --namespace pdb-demo
AI Provider: AI not used; --explain not set

No problems detected
```

4. Both blocked PDBs go unreported. Changing either selector to an equivalent `matchLabels` form makes it appear, which isolates the selector shape as the trigger.

## Environment

- **k8sgpt version:** `v0.4.40-0.20260917124249-8cb270e0800e`, built from `main` at `8cb270e`
- **Kubernetes version:** `v1.34.0` (kind)
- **AI Backend/Provider:** none — reproduces without `--explain`, this is analyzer-side only
- **OS/Platform:** Windows 11

## Additional Context

There is a second, separate defect in the same function worth noting but not conflating: `pdb.go` reads `pdb.Status.Conditions[0]` and assumes `DisruptionAllowed` sits at index 0. Conditions are an unordered set, so the correct lookup is `apimeta.FindStatusCondition`. Today the PDB controller sets only that one condition, so it is latent rather than live.

I have a fix with regression tests ready and will open a PR referencing this issue.

lofoneh:
Closes #1796

## 📑 Description

The PDB analyzer only ever builds a failure inside the loop over `Selector.MatchLabels`, so a `PodDisruptionBudget` with `DisruptionAllowed=False` is silently dropped unless its selector uses `matchLabels`.

Per the `policy/v1` contract:

> A null selector will match no pods, while an empty ({}) selector will select all pods within the namespace.

So two selector shapes went unreported:

- **an empty selector (`{}`)** — applies to every pod in the namespace, the widest blast radius of any PDB
- **a `matchExpressions`-only selector** — ordinary and valid

Both block evictions, which is what wedges a `kubectl drain` and stalls a node upgrade. k8sgpt reported a clean cluster for both.

A **null** selector is deliberately left unreported: it matches no pods, so it cannot block an eviction.

### Reproduction on a real cluster

Two genuinely blocked PDBs on kind (`kubectl get pdb` shows `ALLOWED DISRUPTIONS: 0`, condition `DisruptionAllowed=False / InsufficientPods`):

```yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata: { name: blocks-everything, namespace: pdb-demo }
spec:
  minAvailable: 1
  selector: {}                 # selects EVERY pod in the namespace
---
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata: { name: expr-selector, namespace: pdb-demo }
spec:
  minAvailable: 1
  selector:
    matchExpressions:
      - { key: app, operator: In, values: ["web"] }
```

**Before:**

```
$ k8sgpt analyze --filter PodDisruptionBudget --namespace pdb-demo
No problems detected
```

**After:**

```
$ k8sgpt analyze --filter PodDisruptionBudget --namespace pdb-demo
0: PodDisruptionBudget pdb-demo/blocks-everything()
- Error: InsufficientPods, and its empty selector applies to every pod in namespace pdb-demo

1: PodDisruptionBudget pdb-demo/expr-selector()
- Error: InsufficientPods, expected pdb pod label app in (web)
```

Anonymization is preserved for the new paths: `matchExpressions` keys and values are masked via `util.MaskString`, as is the namespace in the empty-selector message.

## ✅ Checks

## ℹ Additional Information

**An existing test asserted the old behaviour, and this PR changes it.** Calling that out explicitly so it gets reviewed rather than skimmed:

`TestPodDisruptionBudgetAnalyzer` expected a single result from a fixture whose `PDB4` is blocked with an empty selector (commented `// Match Labels Empty`). That expectation encoded the bug, so it now expects both PDBs. It also switches from `results[0].Name` to `require.ElementsMatch`, because results are collected by ranging a map — the ordering was never deterministic, and would have become a flaky test the moment more than one result existed.

**A second, separate defect is deliberately left alone.** This function reads `pdb.Status.Conditions[0]` and assumes `DisruptionAllowed` sits at index 0. Conditions are an unordered set, so the correct lookup is `apimeta.FindStatusCondition`. Only the PDB controller's single condition is set today, so this is latent rather than live — left out to keep this change single-purpose. Happy to follow up with a separate PR if maintainers want it fixed.

**Testing:** `go test ./...` passes. (One unrelated pre-existing failure, `pkg/util TestEnsureDirExists`, reproduces on clean `main` on Windows — it asserts a Unix error string against a Windows one.)


AlexsJones (approved):
Thanks for flagging the changed test expectation in the description instead of letting a reviewer find it, that is the part of this PR that most needed a second pair of eyes and it holds up: `PDB4` in that fixture is a genuinely blocked PDB with an empty selector, so the old `require.Equal(t, 1, len(results))` was asserting the bug. Switching to `require.ElementsMatch` is right for its own reason too, since `preAnalysis` is a map and the ordering was never guaranteed.

Read the whole diff. The guard change from `Selector.MatchLabels != nil` to `Selector != nil` matches the `policy/v1` contract you quoted, and the empty-selector condition (`len(MatchLabels) == 0 && len(MatchExpressions) == 0`) also covers the `matchExpressions: []` spelling, which is likewise an all-pods selector. Nice that it lands on the right answer without special-casing it.

Checked the anonymization path specifically, since that is where this kind of change usually leaks. It is safe: `namespace` appears both in your new empty-selector `Sensitive` entry and in the set `resourceSensitive` derives from `Result.Name`, but `sensitiveMapping` dedupes on the unmasked value and `maskText` replaces longest-first in a single pass, so one identifier cannot end up with two masks or get half-rewritten.

One optional nit, take it or leave it: `strings.ToLower(string(expr.Operator))` renders `DoesNotExist` as `doesnotexist`, so a `DoesNotExist` expression reads "expected pdb pod label app doesnotexist". A small map from operator to a phrase ("in", "not in", "exists", "does not exist") would read better in the AI prompt as well as the terminal. Not worth another round on its own.

Agreed on leaving `Conditions[0]` alone, and thanks for keeping it out rather than bundling it.

CI is green at c099a2e8 (build, golangci-lint, codecov patch and project, validate, DCO). I approved the pending workflow runs so they would actually report. I did not run the branch locally.

approve-recommend: ready for a maintainer's final look and merge.

lofoneh:
Thanks for the thorough review, especially checking the anonymization path — that was the part I was least sure about, so it's good to have the `sensitiveMapping` dedupe and longest-first `maskText` behaviour confirmed rather than assumed.

Good catch on the operator rendering. `NotIn` and `DoesNotExist` come out as "notin" and "doesnotexist", which reads poorly in the terminal and in the prompt. I'll send that as a small follow-up PR with an operator→phrase map rather than another round here, so this one can go in as-is.

Happy to take the `Conditions[0]` → `apimeta.FindStatusCondition` change as a separate follow-up too, if you'd like it.

